PR #122, opened 2023-05-29
Git loader + Antlr in the fly parser

raulraja:
```kotlin
package com.xebia.functional.xef.auto.git

import arrow.core.raise.recover
import com.xebia.functional.xef.antlr.languages.Language
import com.xebia.functional.xef.antlr.languages.antlrParser
import com.xebia.functional.xef.auto.ai
import com.xebia.functional.xef.auto.getOrThrow
import com.xebia.functional.xef.auto.promptMessage
import com.xebia.functional.xef.loaders.FileContent
import com.xebia.functional.xef.loaders.git

suspend fun main() {
  ai {
    contextScope(arrowRepositoryDocuments()) {
      while (true) {
        print("Enter your question: ")
        val line = readlnOrNull() ?: break
        val response = promptMessage(line)
        println(response.firstOrNull())
      }
    }
  }.getOrThrow()
}

private suspend fun arrowRepositoryDocuments() =
  git(
    url = "https://github.com/arrow-kt/arrow",
    fileLoader = ::parseFileWithAntlrIfPossible,
    fileFilter = { it.name.endsWith(".kt") }
  )

private suspend fun parseFileWithAntlrIfPossible(fileContent: FileContent): List<String> {
  val langParser = recover({
    Language.loadByPath(fileContent.path)?.let { language -> antlrParser(language) }
  }) {
    error(it.reason)
  }
  return if (langParser != null) {
    val (_, context, docs) = langParser(fileContent.content)
    if (docs.isNotEmpty()) println("found ${docs.size} docs in ${context::class.java.name}")
    val foundDocuments = docs.map { doc ->
      doc.toDocument(fileContent.path)
    }
    foundDocuments
  } else listOf(fileContent.toDocument())
}
```

The Kotlin parser is defined as:

```kotlin
package com.xebia.functional.xef.antlr.languages

import arrow.core.raise.Raise
import com.xebia.functional.xef.antlr.ANTLRError
import com.xebia.functional.xef.antlr.EntryPoint
import com.xebia.functional.xef.antlr.G4ParserFromURLs

fun Raise<ANTLRError>.kotlin() =
  G4ParserFromURLs(
    EntryPoint(
      method = "kotlinFile",
      lexer = "KotlinLexer",
      parser = "KotlinParser",
      contextDocuments = setOf("KotlinFile", "Declaration"),
    ),
    "https://raw.githubusercontent.com/antlr/grammars-v4/master/kotlin/kotlin/UnicodeClasses.g4",
    "https://raw.githubusercontent.com/antlr/grammars-v4/master/kotlin/kotlin/KotlinLexer.g4",
    "https://raw.githubusercontent.com/antlr/grammars-v4/master/kotlin/kotlin/KotlinParser.g4",
  )
```

This enables a user to dynamically parse sources in repositories for the languages they care about.
We can extend the pattern to add some heuristics and dynamically fetch and create a parser for any existing language using data sources such as https://github.com/antlr/grammars-v4


core/src/commonMain/kotlin/com/xebia/functional/xef/auto/DeserializerLLMAgent.kt
```diff
@@ -39,7 +39,7 @@ suspend inline fun <reified A> AIScope.prompt(
   user: String = "testing",
   echo: Boolean = false,
   n: Int = 1,
-  temperature: Double = 0.0,
+  temperature: Double = 0.5,
```
serras, line 42: Why is this change part of this PR?

examples/kotlin/build.gradle.kts
```diff
@@ -23,11 +23,14 @@ dependencies {
...
+  implementation(projects.xefGit)
+  implementation(projects.xefAntlr)
   implementation(libs.kotlinx.serialization.json)
   implementation(libs.logback)
   implementation(libs.klogging)
   implementation(libs.bundles.arrow)
   implementation(libs.okio)
   implementation(libs.jdbc.mysql.connector)
   api(libs.bundles.ktor.client)
+  implementation("org.antlr:antlr4-runtime:4.13.0")
```
serras, line 35: This should go to the TOML file, I guess.

filesystem/src/commonMain/kotlin/com/xebia/functional/xef/io/FileUtils.kt
```diff
@@ -0,0 +1,48 @@
...
+  source.buffer().use { src ->
+    while (src.read(buffer).also { bytesRead = it } != -1 && !channel.isClosedForWrite) {
+      channel.writeFully(buffer, offset = 0, length = bytesRead)
+      channel.flush()
+    }
+  }
+}
+
+suspend fun readUrlToSink(url: String, sink: Sink) {
+  HttpClient {  }.use {
```
serras, line 38: Note that this is not checking the result status code. If you have a 404, for example, this will give you the body with the error message, which may not be what you want.

```suggestion
  HttpClient { expectSuccess = true }.use {
```

filesystem/src/jsMain/kotlin/com/xebia/functional/xef/io/NodeCommandExecutor.kt
```diff
@@ -15,6 +15,34 @@ actual val CommandExecutor.Companion.DEFAULT: CommandExecutor
   get() = NodeCommandExecutor
 
 object NodeCommandExecutor : CommandExecutor {
+
+  override suspend fun platform(): Platform {
```
serras, line 19: I have two questions regarding this:
1. Why is this needed? At the end of the day this always returns `LINUX(something)`, so we could just set this to `LINUX("node")` and call it a day.
2. The implementation seems very brittle. There are modules in Node which do this, and I think delegating on those would be the right move (if getting the platform name is really required).

serras: Another possibility is to use the [Platform type](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native/-platform/) available in the JS target.

serras (changes requested):

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/DeserializerLLMAgent.kt
raulraja: outdated changes, this branch needs to be merged with `main`. I'll update it.

raulraja:
I'm not sure If I'll have time to update this today, but I'll get to it after the 21st when I get back from vacation. Thanks @serras  for looking into it.

raulraja:
I am closing this to focus on other priorities. In the last couple of months, I learned this is not the best approach, and I plan on rewriting this feature when I have time based on the ReAct agent and a set of git tools.

PR #155, opened 2023-06-01
Generate tokens optimistically matching a pattern

realdavidvega:
In OpenAI models, the available approach to generate completions deterministically, is to guide the model's token generation through biasing. By providing specific `logit_bias` values, we can influence the model's output. For instance, if we want the generation to always move forward, we can request the model to generate one token at a time. However, this method has its limitations, particularly when dealing with longer prompts and completions. For instance, if we aim to generate 1000 tokens (including the prompt), it would require loading and generating a large amount of tokens, which can be very costly.

This small PR aims to optimize this a bit, by doing the following:
- Incorporating the ability of limiting the amount of maximum iterations, e.g. `maxIterations=10`
- Adding the ability of optimistically generating more than 1 token, e.g. `maxTokensPerCompletion=5`
- After generating those tokens, we can see if they partially meet the regex
- Truncate the output until it partially meets the regex, removing not-matching tokens
- Continue iterating reaching the point where the output full matches the regex, or `maxIterations` is reached

By this, we reduce the amount of total calls to OpenAI. However, it would be much more efficient and cheaper to influence the token generation in local models.

cc: @xebia-functional/team-ai


juanpedromoreno (approved):
LGTM, thanks @realdavidvega !

PR #156, opened 2023-06-01
Improve publish workflow

franciscodr:
This pull request proposes two improvements to the publishing workflow:

- It generates the Scala documentation by running the `scaladoc` task provided by the Scala Gradle plugin. 
- It updates the workflow that publishes final versions to be triggered after creating a new final version tag.


juanpedromoreno (approved):
Thanks @franciscodr !

PR #157, opened 2023-06-01
Scala prompt templating

necosta:
Following on https://github.com/xebia-functional/xef/pull/143 comments, perhaps a more useful abstraction/templating could be done for a chain of prompts where one prompt response feeds the next prompt.

I have this initial version setup with an example included, let me know your thoughts on this.

### DIFF
--- a/scala/src/main/scala/com/xebia/functional/xef/scala/prompt/PromptTemplate.scala
+++ b/scala/src/main/scala/com/xebia/functional/xef/scala/prompt/PromptTemplate.scala
@@ -5,10 +5,10 @@ import io.circe.Decoder
 
 trait PromptTemplate[A: ScalaSerialDescriptor: Decoder] {
 
-  def create(template: String): A
+  def chain(template: String): A
 
   extension (a: A) {
-    def add[B: ScalaSerialDescriptor: Decoder](template: A => String): B =
+    def chain[B: ScalaSerialDescriptor: Decoder](template: A => String): B =
       ai(prompt[B](template(a)))
   }
 }
@@ -18,4 +18,4 @@ object PromptTemplate:
   def apply[A](using ev: PromptTemplate[A]): PromptTemplate[A] = ev
 
   inline final def derived[A: ScalaSerialDescriptor: Decoder]: PromptTemplate[A] = new PromptTemplate[A]:
-    def create(template: String): A = ai(prompt[A](template))
+    def chain(template: String): A = ai(prompt[A](template))
--- a/scala/src/main/scala/com/xebia/functional/xef/scala/prompt/PromptTemplate.scala
+++ b/scala/src/main/scala/com/xebia/functional/xef/scala/prompt/PromptTemplate.scala
@@ -1,0 +1,21 @@
+package com.xebia.functional.xef.scala.prompt
+
+import com.xebia.functional.xef.scala.auto.*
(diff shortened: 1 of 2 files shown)


necosta:
Rebased and improved the PR. Ready for review.

juanpedromoreno (approved):
LGTM, thanks @necosta !

scala/src/main/scala/com/xebia/functional/xef/scala/prompt/PromptTemplate.scala
```diff
@@ -0,0 +1,22 @@
...
+import com.xebia.functional.xef.scala.auto.*
+import io.circe.Decoder
+
+trait PromptTemplate[A] {
+
+  def chain(template: String): A
+
+  extension (a: A) {
+    def chain[B: ScalaSerialDescriptor: Decoder](template: A => String): B =
+      ai(prompt[B](template(a)))
```
Yawolf, line 12: I'm not very sure if this should live here or not, but what do you think about adding a infix operator? Something like `>>` so we can do this:
```scala
PromptTemplate[Play](Play("The power of Zuluastral", "Modern Era")) >> synopsisTemplate >> reviewTemplate >> scoreTemplate
```
Is it possible? Does it make sense? If not, just forget it! 😄

Yawolf (approved):

scala/src/main/scala/com/xebia/functional/xef/scala/prompt/PromptTemplate.scala
necosta: I think this syntactic sugar is possible yes but would need to try it. 
I'll take a note, will do it in a future PR if it does work.

scala/src/main/scala/com/xebia/functional/xef/scala/prompt/PromptTemplate.scala
```diff
@@ -0,0 +1,22 @@
+package com.xebia.functional.xef.scala.prompt
+
+import com.xebia.functional.xef.scala.auto.*
+import io.circe.Decoder
+
+trait PromptTemplate[A] {
+
+  def chain(template: String): A
```
raulraja, line 8: This functionality should be developed in Kotlin multiplatform and mapped to other lang bindings. @juanpedromoreno @necosta  What criteria are we following to decide what gets developed in Kotlin vs. Scala?

juanpedromoreno: Hello @raulraja, I agree with your observation. Nelson and I have had an offline discussion regarding this matter. The Scala API was unable to create prompts beyond utilizing string interpolations, unlike Kotlin does [here](https://github.com/xebia-functional/xef/blob/bd7e3bc8e8088682889e000d0c69ba9b025f2d1d/core/src/commonMain/kotlin/com/xebia/functional/xef/prompt/Prompt.kt). Due to this limitation, we ended up creating different designs without considering the possible divergence. I suggest we work on unifying the APIs once again 👍

necosta: Will try to move this logic to Kotlin. Scala should follow the Kotlin API yes, it was just overlooked.

PR #159, opened 2023-06-02
ScalaSerialDescriptor -> SerialDescriptor

Yawolf:
- Now `ScalaSerialDescriptor` is called just `SerialDescriptor`
- `ScalaSerialDescriptorContext` has been renamed to `SerialDescriptorInstances`, and now is a class from which extends the `SerialDescriptor` companion object in order to avoid given imports.
- Docs updated
- Added two `TODO` regarding the undefined implementation of `serialize` and `deserialize` methods in `KSerializer` instance.


scala/src/main/scala/com/xebia/functional/xef/scala/auto/SerialDescriptorContext.scala
```diff
@@ -0,0 +1,64 @@
+package com.xebia.functional.xef.scala.auto
```
necosta, line 1: Rename file?

necosta (commented):
I think you have to rebase as merging does not fix the issues on file SynopsisReview and PromptTemplate already merged into main.

necosta:
The changes look good, happy to approve after rename conflicts get solved.

scala/src/main/scala/com/xebia/functional/xef/scala/auto/SerialDescriptorContext.scala
Yawolf: Oh, right!

Yawolf: Done!

necosta (approved):

juanpedromoreno (approved):

juanpedromoreno (approved):

PR #166, opened 2023-06-05
First approach of loading GPT4All models locally

realdavidvega:
Creation of `gpt4all-kotlin` gradle project including a first version of loading GPT4All models locally.

For now only Kotlin/JVM target is included, using JNA for the native bindings (for `darwin-aarch64`, but other platforms can be included), but we could potentially make it available to other platforms.


juanpedromoreno (approved):
LGTM overall, great work @realdavidvega !

examples/kotlin/src/main/kotlin/com/xebia/functional/xef/auto/gpt4all/Chat.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.xebia.functional.gpt4all.Message
+import com.xebia.functional.gpt4all.Response
+import com.xebia.functional.xef.AIError
+import com.xebia.functional.xef.auto.ai
+import com.xebia.functional.xef.auto.getOrThrow
+import java.nio.file.Path
+import java.util.*
+
+suspend fun main() {
+    ai {
```
carbaj03, line 15: Here ai {} is just used like an effect {}. In other words, you aren't using the ai {} for prompting or something LLM related. This is a different approach than the one used with OpenAi and Hugginface

realdavidvega: Right, the idea is to give the option of having it as part of the AIScope in the near future, but to avoid confusion I will just wrap it on an `either` block for now.

realdavidvega:
Ready to review. For now I think we can review it in its current state.

Before integrating with `AIScope` we should have the embeddings part ready, which implies creating more native bindings for LLAMA.cpp. Both tasks can be discussed and implemented in upcoming PRs.

cc/ @xebia-functional/team-ai

necosta (dismissed):
I got the example running locally, with the small caveat of having to rename resources folder `darwin-aarch64` to `darwin-x86-64`.
To me it is ready to be merged, great job!

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/GPT4All.kt
```diff
@@ -0,0 +1,86 @@
...
+        operator fun invoke(
+            path: Path,
+            modelType: LLModel.Type
+        ): GPT4All = object : GPT4All {
+            val gpt4allModel: GPT4AllModel = GPT4AllModel(path, modelType)
+
+            override suspend fun createCompletion(request: CompletionRequest): CompletionResponse =
+                with(request) {
+                    val response: String = generateCompletion(prompt, generationConfig)
+                    val name: String = gpt4allModel.llModel.name
```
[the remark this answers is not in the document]
realdavidvega: Thanks @necosta! I've included the `darwin-x86-64` platform also. And applied the suggestion 😃

juanpedromoreno (approved):

necosta (approved):

PR #169, opened 2023-06-07
Create plugin for Scala publishing conventions

franciscodr:
This pull request creates a custom plugin to gather all the settings to publish Scala code to Maven Central. It includes:

- Generate Javadoc jar from the output of the `scaladoc` task
- Generate POM file
- Sign the artifacts


juanpedromoreno (approved):
Awesome, thanks @franciscodr !

buildSrc/src/main/kotlin/ScalaPublishingConventionsPlugin.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.gradle.api.tasks.bundling.Jar
+import org.gradle.kotlin.dsl.get
+import org.gradle.kotlin.dsl.register
+import org.gradle.plugins.signing.SigningExtension
+
+class ScalaPublishingConventionsPlugin : Plugin<Project> {
+  override fun apply(project: Project): Unit = project.run {
+    val scaladocJarTask: TaskProvider<Jar> = tasks.register<Jar>("scaladocJar") {
+      group = BasePlugin.BUILD_GROUP
+      tasks.findByName("scaladoc")?.let { dependsOn(it) }
```
nomisRev, line 18: What if `scaladoc` is not available? 🤔 Does this task still do anything?

Alternatively, we can only add `scaladocJar` if `scaladoc` is available or I would use `requireNotNull` with a meaningful message here. (I am not entirely sure if we should throw `IllegalStateExeption` in Gradle though, I think we need to use the proper reporting tools).

nomisRev (dismissed):
Looks great @franciscodr 🙌 Just one question/doubt.

buildSrc/src/main/kotlin/ScalaPublishingConventionsPlugin.kt
franciscodr: Good point. I guess we should control all the required settings. For instance, the signing key is mandatory when publishing to Maven Central. Let me think about it 🤔

raulraja (approved):

nomisRev (approved):

Yawolf (approved):

PR #170, opened 2023-06-07
Cu 861mwhnjx split kotlinx serialization

nomisRev:
This PR creates the `xef-kotlin` module, which will serve as the new top-level module for Kotlin. Where `xef-core` is going to be reduced in size, and APIs `xef-kotlin` will cover the use-cases we have specifically for Kotlin.

The goal is that `xef-core` **only contains** Java (& Scala) friendly APIs, and thus definitions such as `context(AIScope)` and `typealias AI<A> = context(AIScope) () -> A` are meant for `xef-kotlin`.


nomisRev:
> Cancelled after 15m

Is the build running for longer than 15m 🤕

Yawolf (dismissed):
This is great!

Yawolf (approved):

PR #174, opened 2023-06-07
Redefines the Gradle build strategy

juanpedromoreno:
This PR could be considered a WIP, even if it's finally merged. Further iterations and optimizations will be required.

Initially, the objective of this PR is to decrease the build time. For that reason:

* I'm creating a pipeline matrix with the different platforms: `jvm`, `js`, `linuxX64`, `macosX64`, `mingwX64` (notice I've removed `macosArm64` due to some incompatibilities).
* I split the constructions and tests into _multi-platform_ and _single-platform_ (JVM). This could be done by creating the groups at the Gradle level, which could be the subject of further improvements.
* Regarding JS, I'm not building for browser and node, which might be required for certain modules, but I left that out of this work.

## Result

Hopefully, the build times should be reduced by more than 10 minutes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
 
+    strategy:
+      matrix:
+        platform: [ jvm, js, linuxX64, macosX64, mingwX64 ]
+
     steps:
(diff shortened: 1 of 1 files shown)


juanpedromoreno:
Supersed by https://github.com/xebia-functional/xef/pull/175

PR #175, opened 2023-06-07
Redefines the Gradle build strategy

juanpedromoreno:
This PR could be considered a WIP, even if it's finally merged. Further iterations and optimizations will be required.

Initially, the objective of this PR is to decrease the build time. For that reason:

* I'm creating a pipeline matrix with the different platforms: `jvm`, `js`, `linuxX64`, `macosX64`, `mingwX64` (notice I've removed `macosArm64` due to some incompatibilities).
* I split the constructions and tests into _multi-platform_ and _single-platform_ (JVM). This could be done by creating the groups at the Gradle level, which could be the subject of further improvements.
* Regarding JS, I'm not building for browser and node, which might be required for certain modules, but I left that out of this work.

## Result

Hopefully, the build times should be reduced by more than 10 minutes.


.github/workflows/build.yml
```diff
@@ -23,10 +27,29 @@ jobs:
...
-      - name: build
+      - name: Setup Gradle
         uses: gradle/gradle-build-action@v2
-        with:
-          arguments: build --full-stacktrace
+
+      - name: Build Multi-platform Modules
+        run: |
+          ./gradlew \
+            :xef-core:${{ matrix.platform }}Test \
```
nomisRev, line 36: We're excluding some other checks like formatting with this check.

.github/workflows/build.yml
```diff
@@ -12,6 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
 
+    strategy:
+      matrix:
+        platform: [ jvm, js, linuxX64, macosX64, mingwX64 ]
```
nomisRev, line 17: We cannot build `macosX64` or `mingwX64` on ubuntu? 🤔

juanpedromoreno: Done, please @nomisRev , check whenever you have time

.github/workflows/build.yml
juanpedromoreno: Added to the pipeline

Yawolf (commented):
Is this still a WIP PR? It would be very beneficial to merge this asap 🙂

diesalbla:
@Yawolf Is this compatible with the #197 , or would that take it far apart?

raulraja:
@juanpedromoreno do we still need this, what can we do to get it merged?

juanpedromoreno:
I think @franciscodr and @nomisRev were testing a different approach. I don't want to make it conflict with theirs, but I agree with you; let's decide to speed up the builds and go for one. Happy to close this one if it doesn't make sense 👍

serras (approved):
Nice trick to handle for multi- and single-platform targets!

raulraja:
@franciscodr @nomisRev can this be merged?
@juanpedromoreno I see a hardcoded list of modules there assumed to be multiplatform. Can we get this dynamically from the project build? When adding new modules, I see how this may not work as intended if they are not added to this list.

juanpedromoreno:
@Yawolf was going to look at this. Please, Yago, look also at that particular comment from @raulraja . Thanks

> I see a hardcoded list of modules there assumed to be multiplatform. Can we get this dynamically from the project build?

Yawolf:
> @Yawolf was going to look at this. Please, Yago, look also at that particular comment from @raulraja . Thanks
> 
> > I see a hardcoded list of modules there assumed to be multiplatform. Can we get this dynamically from the project build?

@raulraja I changed that by doing this:
https://github.com/xebia-functional/xef/pull/175/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R18

Let me know if that makes sense to you 🙂 

So far, `./gradlew build` takes around 7m:
```
BUILD SUCCESSFUL in 7m 36s
275 actionable tasks: 263 executed, 12 up-to-date
```

juanpedromoreno:
Right,

Maybe something like this @Yawolf ?

```kotlin
fun Project.isMultiplatformModule(): Boolean {
    val kotlinPluginId = "org.jetbrains.kotlin.multiplatform"
    return plugins.hasPlugin(kotlinPluginId)
}

val multiPlatformModules = project.subprojects.filter { it.isMultiplatformModule() }
```

juanpedromoreno:
👍 Great, that list looks correct, yes.

Yawolf:
In my local, with these changes, it takes:
```
BUILD SUCCESSFUL in 6m 39s
264 actionable tasks: 250 executed, 14 up-to-date
```

While the main branch takes:
```
BUILD SUCCESSFUL in 6m 44s
264 actionable tasks: 250 executed, 14 up-to-date
```

In GitHub actions, this PR took 7m 36s to build, while PR #130 took  8m 59s. So there's a slight performance improvement in these changes.

@raulraja @juanpedromoreno

juanpedromoreno:
Sounds good, thanks @Yawolf

Yawolf (approved):

nomisRev (approved):

PR #177, opened 2023-06-08
Remove Arrow Resilience Schedule

nomisRev:
This PR removes the dependency on Arrow Resilience Schedule as part of the work to remove Arrow from the _core_ library, and it instead simply relies on the retry mechanism of Ktor.

I've also removed the `RetryConfig`, as I assume we're not using this yet and we can instead simply rely on sane defaults until otherwise is needed. @Montagon @javipacheco are you using a custom `RetryConfig` in your project?


core/src/commonMain/kotlin/com/xebia/functional/xef/ktor.kt
```diff
@@ -21,15 +20,14 @@ inline fun <reified A> HttpRequestBuilder.configure(token: String, request: A):
...
 }
 
-suspend fun ResourceScope.httpClient(engine: HttpClientEngine?, baseUrl: Url): HttpClient =
+suspend fun ResourceScope.httpClient(baseUrl: Url): HttpClient =
   install({
-    engine?.let { HttpClient(engine) { configure(baseUrl) } } ?: HttpClient { configure(baseUrl) }
+    HttpClient {
+      install(HttpTimeout)
+      install(ContentNegotiation) { json() }
+      install(HttpRequestRetry)
```
franciscodr, line 28: Are we using the default retry configuration provided by `ktor`?

franciscodr (approved):
Thanks, @nomisRev! Just a question

core/src/commonMain/kotlin/com/xebia/functional/xef/ktor.kt
franciscodr: Nevermind, I saw you comment in the description

PR #178, opened 2023-06-08
Remove Arrow Resource from Core

nomisRev:
This PR removes `ResourceScope` from the app, relying mostly on `AutoCloseable` which recently added to the Kotlin Multiplatform API.

This removes the ability to _install_ resources in the `AIScope`, but we can in the future still support this by implementing an `AutotCloseable` DSL (for Scala and Java) too.


core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AI.kt
```diff
@@ -145,23 +136,16 @@ class AIScope(
    * within [block] will be finalized after [block] finishes.
    */
```
franciscodr, line 137: We should update this piece of documentation by removing the `resourceScope` reference.
```
/**
   * Creates a new scoped [VectorStore] using [store], which is scoped to the [block] lambda. The
   * [block] also runs on a _nested_ [resourceScope], meaning that all additional resources created
   * within [block] will be finalized after [block] finishes.
   */
```

franciscodr (approved):
LGTM 🥳

juanpedromoreno (approved):

Yawolf (approved):

PR #179, opened 2023-06-12
Deleting Arrow STM

Yawolf:
Following this ticket, in this PR we are deleting `Arrow SML` in favor of `Atomic` + `State`, which is a product (data class) with the `LocalVectorStore` state, i.e.: `documents` and `precomputedEmbeddings`


nomisRev (approved):
Looks great @Yawolf 🙌

PR #186, opened 2023-06-14
Gpt 3.5 turbo 0613 serialization support

raulraja:
Support for Open AI function calling on chat completions:
https://platform.openai.com/docs/guides/gpt/function-calling

This means we no longer need to instruct the LLM to produce serializable json based on a schema.
Instead, we create a function whose parameters are the schema per the new Open AI support for function calls on.
For now available in `gpt-3.5-turbo-0613` and missing for `gpt-4-0613`, which I have not been able to test but should be a matter of adding the constants in the model list.


raulraja:
@xebia-functional/team-ai

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/functions/CFunction.kt
```diff
@@ -0,0 +1,7 @@
+package com.xebia.functional.xef.llm.openai.functions
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonObject
+
+@Serializable
+data class CFunction(val name: String, val description: String, val parameters: JsonObject)
```
Yawolf, line 7: Aren't you adding here again what we are trying to avoid? We don't want to depend on KotlinX in Scala or Java. Suppose the field `parameters` is a KotlinX `JsonObject`. In that case, we will need to use KotlinX in Scala or Java, and we don't want that... maybe a `jsonStrParameter: String` is more generic and easy to port to other languages?

Yawolf: We want `xef-core` to be the most agnostic possible on Arrow and KotlinX. A new module was created called xef-kotlin, which will use Arrow and KotlinX. So do you think this parameter can be set as a `String` so `xef-core` does not depend on `KotlinX` again?

raulraja: > Aren't you adding here again what we are trying to avoid? We don't want to depend on KotlinX in Scala or Java. Suppose the field `parameters` is a KotlinX `JsonObject`. In that case, we will need to use KotlinX in Scala or Java, and we don't want that... maybe a `jsonStrParameter: String` is more generic and easy to port to other languages?

No, you should never have to access `parameters` from Scala or elsewhere since constructing that it's handled internally by the `prompt` functions in the core library. You are calling those functions from Scala, but you never have to access or use these parameters internals. The endpoint call and new Open AI spec require `parameters` to be a dynamic JSON Schema for which you can't even construct a sane type and serialize in Scala. This is OpenAI specific and all these models should be moved outside of core once core is generic for any LLM but the fact that `parameters` is a Kotlin JsonObject should not affect Scala or any other integration.

> We want `xef-core` to be the most agnostic possible on Arrow and KotlinX. A new module was created called xef-kotlin, which will use Arrow and KotlinX. So do you think this parameter can be set as a `String` so `xef-core` does not depend on `KotlinX` again?

This is specific to the OpenAI client, it will end up wherever that is moved out of the core, probably another module.
Generating the client automatically right now from the official Open AI spec is challenging because the Open API generator for Kotlin multiplatform does not support `oneOf` or json-schema as inner properties.

https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L2398-L2409

https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L2337-L2355

Yawolf: >No, you should never have to access parameters from Scala or elsewhere since constructing that it's handled internally by the prompt functions in the core library.

Probably I'm misunderstanding something, but I see `promptMessage`, which is an AiDsl member, requires a `List<CFunction>`, and `CFunction` has the `parameter: JsonObject` filed. So, if I want to use `promptMessage`, I need to provide a list of `CFunction`, no? And for generating a `List[CFunction]` in Scala/Java I'd need to generate a `parameter` field with type `CFunction`, no?

I'm missing something? 🤔

raulraja: The list of `CFunction` is generated inside the core in `encodeFunctionSchema`
https://github.com/xebia-functional/xef/blob/be93241a6c44e3171e79451b81aaaacb4b8c2ed7/kotlin/src/commonMain/kotlin/com/xebia/functional/xef/auto/serialization/functions/FunctionSchema.kt#L30-L32
which we are calling from `Scala` using the serial descriptor:
https://github.com/xebia-functional/xef/blob/be93241a6c44e3171e79451b81aaaacb4b8c2ed7/scala/src/main/scala/com/xebia/functional/xef/scala/auto/package.scala#L52-L69

You don't construct a CFunction yourself and CFunction would never be exposed to the Scala API as a public type.

Yawolf: Oookay, gotcha, thanks! 😄

raulraja: Yes, we plan on getting rid of all the Open-AI specific parameters, you should be able to specify the model, and the rests are dependent on it and computed internally. Still, I did not want to make this a more extensive PR.

In particular, the only LLMs that support functions like this are Open-AI, but `promptMessage` is supposed to work with more. The functions params and others specific to Open-AI should be taken as a single config argument for the prompt, which would be set to the AIScope default config params.

I have to focus on other things for the rest of the day, and I'd rather address all Open AI concerns in a single PR when we make `core` generic, but if you want to play with it or change the Scala parts, feel free to do it in this same branch or as a PR to this one.

Yawolf: Thanks Raul! 😄

Yawolf (approved):

raulraja:
@franciscodr @realdavidvega you were interested in this area. I'd like to use this work in the demo @franciscodr @nomisRev and @franciscodr are working on. If you have a chance, please let me know if this needs changes before we merge it 🙏

PR #187, opened 2023-06-16
Extension Methods to Instance Methods

Yawolf:
In this PR we are getting rid of all extension methods for `AIScope` in `xef-core`, and defining them as instance methods, so it can provide a more straightforward and understandable API for language integration.

Previously, something like this was needed:
```scala
def foo(using scope: AIScope) = KotlinAIScope.foo(scope.kt)
```
Now it's more simple:
```scala
def foo(using scope: AIScope) = scope.foo
```


core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIScope.kt
```diff
@@ -0,0 +1,501 @@
...
+
+  /**
+   * Creates a nested scope that combines the provided [store] with the outer _store_. This is done
+   * using [CombinedVectorStore].
+   *
+   * **Note:** if the implementation of [VectorStore] is relying on resources you're manually
+   * responsible for closing any potential resources.
+   */
+  @AiDsl
+  suspend fun <A> contextScope(store: VectorStore, block: AI<A>): A =
```
nomisRev, line 81: This probably another candidate for the `AsFuture` plugin 🤔 Where the `block` argument needs to be translated from `suspend AIScope.() -> A` to `AIScope.() -> Future<A>`.

`AIScope.() -> Future<A>` will appear as `(AIScope) -> Future<A>` from Java.

nomisRev (approved):
Looks good @Yawolf! Thank you

PR #193, opened 2023-06-20
Paving the Way for Model Flexibility in xef.core and Increasing GPT-3.5 Turbo's Context Window

raulraja:
# Summary
This PR marks an important shift in the xef.core module, as it lays the groundwork for removing the dependency on OpenAI and enabling the usage of local models and other types of models within the ai block. This flexibility allows the system to be more adaptive and versatile in accommodating various AI models.

Moreover, this PR also includes an update to increase the context window of the GPT-3.5 Turbo model to 16k tokens, enabling the model to consider a larger context when generating responses, thereby potentially improving the model's understanding of the provided context and its ability to generate more relevant responses.

# Details
The changes introduced in this PR include:

- Refactoring of AI.kt to use AIClient instances instead of specifically OpenAIClient instances. This allows for more versatility and flexibility in the types of AI models that can be used within the ai block.

- Creation of a new file AIRuntime.kt that defines an AIRuntime data class. This class provides a structured way to override the runtime environment for executing AI blocks.

- Changes to AIScope.kt to accommodate the updates in AI.kt and AIRuntime.kt. The updates include the addition of new parameters like `defaultModel`, `defaultSerializationModel`, and `AIClient`, among others.

- The context window of the GPT-3.5 Turbo model has been increased to 16k tokens in the AIScope of the default Open AI AIRuntime.kt. This update allows the model to consider a larger context when generating responses, potentially improving its understanding of the documents in context and its ability to generate more relevant responses. The documents in context default are now 20.


raulraja:
@xebia-functional/team-ai

rafaparadela (approved):

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.xebia.functional.xef.llm.openai.LLMModel
+import com.xebia.functional.xef.vectorstores.LocalVectorStore
+import kotlin.jvm.JvmStatic
+import kotlin.time.ExperimentalTime
+
+data class AIRuntime<A>(val runtime: suspend (block: AI<A>) -> A) {
+  companion object {
+    @OptIn(ExperimentalTime::class)
+    @JvmStatic
+    fun <A> openAI(): AIRuntime<A> = AIRuntime { block ->
```
franciscodr, line 15: I guess this is something that we will move to a specific module for OpenAI so we don't include a dependency to OpenAI in the core library, won't we?

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AI.kt
```diff
@@ -58,7 +56,14 @@ suspend fun <A> MockAIScope(
   try {
     val embeddings = OpenAIEmbeddings(OpenAIConfig(), mockClient)
     val vectorStore = LocalVectorStore(embeddings)
-    val scope = AIScope(mockClient, vectorStore, embeddings)
+    val scope =
```
franciscodr, line 59: Not to address in this pull request, but we mentioned moving out the `MockAIScope ` to a separate `test` module

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
```diff
@@ -0,0 +1,32 @@
...
+      val openAIConfig = OpenAIConfig()
+      KtorOpenAIClient(openAIConfig).use { openAiClient ->
```
franciscodr, line 17: We need to redesign this part to prevent creating an instance of the HTTP client for every `ai` block

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AI.kt
```diff
@@ -6,7 +6,10 @@ import arrow.core.right
 import com.xebia.functional.xef.AIError
 import com.xebia.functional.xef.embeddings.OpenAIEmbeddings
 import com.xebia.functional.xef.env.OpenAIConfig
-import com.xebia.functional.xef.llm.openai.*
+import com.xebia.functional.xef.llm.openai.AIClient
```
franciscodr, line 9: `AIClient` should be located in a different package because it's unrelated to `openai`, isn't it?

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
```diff
@@ -0,0 +1,32 @@
...
+        val embeddings = OpenAIEmbeddings(openAIConfig, openAiClient)
+        val vectorStore = LocalVectorStore(embeddings)
+        val scope =
+          AIScope(
+            defaultModel = LLMModel.GPT_3_5_TURBO_16K,
+            defaultSerializationModel = LLMModel.GPT_3_5_TURBO_FUNCTIONS,
```
franciscodr, line 23: Are the serialization models generic to all the platforms? Or are they only specific to OpenAI?

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
raulraja: Yes, this PR does not address moving open ai deps out of the core. We still need to move the open ai dependencies out of the core. It's just the first step toward that, trying to cause as few file movements as possible. We plan on using the open ai models as a baseline interface for the rest of the models to adapt to them.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AI.kt
raulraja: Once we have a local model in the ai DSL, we should not have these mocks but true integration tests against our model of choice.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
raulraja: We will get rid of our custom Ktor implementation in favor of https://github.com/aallam/openai-kotlin, which just added function support. All our client-related code for open ai will then be gone. that lib keeps up with the entire API, including the missing methods for images and others.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AI.kt
raulraja: yes, when we move Open ai out of core and depend on https://github.com/aallam/openai-kotlin we will remove the openai package

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
raulraja: No, we expect that for local models we will have to implement the function interface similar to OpenAI, desugaring the request to whatever serialization mechanism we can use with a local model that already specs json from a schema.
We also plan to split the types of the features of the client so they are more in the lines of Audio, Video, Text, Chat, etc.. and these are typed so you are unable to run a `prompt` or `image` in models where it does not make sense. We still need to implement all this.

PR #195, opened 2023-06-21
Local-Vector-Stores / addStores: make one update.

diesalbla:
Instead of having a sequence of updates to the Atomic State, it may be more efficient to have a single bulk update. It would also be more transactional.

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
@@ -35,8 +35,8 @@ private constructor(private val embeddings: Embeddings, private val state: Atomi
     val embeddingsList =
       embeddings.embedDocuments(texts, chunkSize = null, requestConfig = requestConfig)
     state.getAndUpdate { prevState ->
-      val newEmbeddings = prevState.precomputedEmbeddings ++ texts.zip(embeddingsList)
-      State(prevState.documents ++ texts, newEmbeddings)
+      val newEmbeddings = prevState.precomputedEmbeddings + texts.zip(embeddingsList)
+      State(prevState.documents + texts, newEmbeddings)
     }
   }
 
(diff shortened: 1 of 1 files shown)


nomisRev (dismissed):

core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
```diff
@@ -34,13 +34,9 @@ private constructor(private val embeddings: Embeddings, private val state: Atomi
...
       embeddings.embedDocuments(texts, chunkSize = null, requestConfig = requestConfig)
-    texts.zip(embeddingsList) { text, embedding ->
-      state.getAndUpdate { prevState ->
-        State(
-          documents = prevState.documents + text,
-          prevState.precomputedEmbeddings + Pair(text, embedding)
-        )
-      }
+    state.getAndUpdate { prevState ->
+      val newEmbeddings = prevState.precomputedEmbeddings ++ texts.zip(embeddingsList)
```
nomisRev, line 38:
```suggestion
      val newEmbeddings = prevState.precomputedEmbeddings + texts.zip(embeddingsList)
```

core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
```diff
@@ -34,13 +34,9 @@ private constructor(private val embeddings: Embeddings, private val state: Atomi
...
+      State(prevState.documents ++ texts, newEmbeddings)
```
nomisRev, line 39:
```suggestion
      val newEmbeddings = prevState.precomputedEmbeddings + texts.zip(embeddingsList)
      State(prevState.documents + texts, newEmbeddings)
```

In Kotlin we simply use `+` for _appending_.

franciscodr (approved):

franciscodr (approved):

PR #196, opened 2023-06-21
Generic AI client and models + open-ai client lib

raulraja:
Delegates the open ai client runtime to https://github.com/aallam/ and reuses previous models as generic classes for xef core. 

Adds a new hierarchy for LLM Models to be typed based on their capabilities. This prevents sending the wrong kind of model to an operation that does not support that capability.

For example the `prompt` function can only be invoked with a model that has serialization capabilities through functions.
```kotlin
@AiDsl
suspend inline fun <reified A> AIScope.prompt(
  question: String,
  ...
  model: LLM.ChatWithFunctions = this.defaultSerializationModel
  ...
): A
```

I plan to eliminate the AIClient in the next PR and implement the capabilities in the models directly to remove the client dependency from the scope. The client should not be a scoped object because each call can be to different models with different client implementations inside the AI block.


raulraja:
@xebia-functional/team-ai

diesalbla:
Would that be https://github.com/aallam/openai-kotlin?

diesalbla (dismissed):
To my limited understanding, seems ok...

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
```diff
@@ -13,9 +15,14 @@ data class AIRuntime<A>(val runtime: suspend (block: AI<A>) -> A) {
     @OptIn(ExperimentalTime::class)
     @JvmStatic
     fun <A> openAI(): AIRuntime<A> = AIRuntime { block ->
-      val openAIConfig = OpenAIConfig()
-      KtorOpenAIClient(openAIConfig).use { openAiClient ->
-        val embeddings = OpenAIEmbeddings(openAIConfig, openAiClient)
+      val openAIConfig =
+        OpenAIConfig(
```
rafaparadela, line 19: Now we are using the `OpenAIConfig` from `com.aallam.openai` I guess we can delete [our domain's](https://github.com/xebia-functional/xef/blob/open-ai-typed-models/core/src/commonMain/kotlin/com/xebia/functional/xef/env/config.kt#L5):

rafaparadela (approved):
Besides a minor comment, the changes look good to me.

Note: I'm wondering if OpenAI-Kotlin implements `Embeddings` in a way that we can use in `OpenAIEmbeddings`.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
```diff
@@ -1,9 +1,11 @@
 package com.xebia.functional.xef.auto
 
-import com.xebia.functional.xef.embeddings.OpenAIEmbeddings
-import com.xebia.functional.xef.env.OpenAIConfig
-import com.xebia.functional.xef.llm.openai.KtorOpenAIClient
-import com.xebia.functional.xef.llm.openai.LLMModel
+import com.aallam.openai.client.OpenAI
+import com.aallam.openai.client.OpenAIConfig
```
nomisRev, line 4: I thought the goal of the new `openai` module, and this effort was to abstract away OpenAI from `core`.
So we shouldn't be relying directly on `openai` from `core`, but instead only on `AIClient`.

So this `openAI()` smart-constructor` should be defined in the `openai` module, rather than here.

Are we directly relying on this in Core after all? I thought the idea was to split this off. We should be relying on AIClient`, so this `openAI` function should be defined in the `openai` module

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
```diff
@@ -13,9 +15,14 @@ data class AIRuntime<A>(val runtime: suspend (block: AI<A>) -> A) {
...
-      val openAIConfig = OpenAIConfig()
-      KtorOpenAIClient(openAIConfig).use { openAiClient ->
-        val embeddings = OpenAIEmbeddings(openAIConfig, openAiClient)
+      val openAIConfig =
+        OpenAIConfig(
+          token =
+            requireNotNull(getenv("OPENAI_TOKEN")) { "OpenAI Token missing from environment." },
+        )
+      val openAI = OpenAI(openAIConfig)
+      OpenAIClient(openAI).use { openAiClient ->
```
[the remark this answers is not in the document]
nomisRev: This should probably be parameterised in the constructor of `openAI`, since not everyone might want to use environment variables for this. Perhaps it's coming from a `application.conf` instead.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/OpenAIClient.kt
```diff
@@ -0,0 +1,255 @@
+package com.xebia.functional.xef.llm.openai
+
+import com.aallam.openai.api.BetaOpenAI
```
nomisRev, line 3: Is this still being moved to a different module in a subsequent PR?

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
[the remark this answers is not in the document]
raulraja: I parameterized the config as a nullable argument so you can pass it through or, if missing, attempts to load it from the environment in https://github.com/xebia-functional/xef/pull/196/commits/dd47c872288c2b1d23393e5a248f7d146f6071a4

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/chat/ChatCompletionRequestWithFunctions.kt
```diff
@@ -0,0 +1,20 @@
+package com.xebia.functional.xef.llm.models.chat
+
+import com.xebia.functional.xef.llm.models.functions.CFunction
+
+data class ChatCompletionRequestWithFunctions(
```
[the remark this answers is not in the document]
raulraja: In practice for the Open AI use case, ChatCompletion also deals with functions at the expense of many loose parameters that the remote API interprets one way or another based on which ones you set from the request.
I found this more complex to understand than having two features, chat and chat with functions, and that is why it's modeled as two independent objects. For more context:
https://platform.openai.com/docs/api-reference/chat/create#chat/create-functions
https://platform.openai.com/docs/api-reference/chat/create#chat/create-function_call

All those fields and the returns of the messages in the response get more complex to handle in the client when it's all in one object that can do multiple things.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
raulraja: That would be ideal, but based on our conversation on Slack yesterday I thought we still need the openAI deps in core if we want to provide a single module with sane defaults, and those defaults are the OpenAI ones.

Alternatively, you would have to bring a `bom` or two modules, xef core for the base, and the OpenAI runtime. Things like `getOrElse` may have to live in the OpenAI module because that operation decides what runtime it uses.

I think the AIClient over the scope is a bad abstraction because the client is dependent on each operation and may change with each operation depending on the model you use not the scope. I plan on completely getting read of the `AIClient` and trying to move those specific operations like `createChatCompletions` or `createImage` to the actual model hierarchy so chat and image models independently implement those.

Still, if we want xef-core to run independently without additional deps we need the OpenAI dep here or a different strategy.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/OpenAIClient.kt
raulraja: Would that new module be a dependency of core? if that is the case we can move the remaining 3 files for the OpenAI impl to a new module and link it.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
raulraja: Removed in [76d0e9e](https://github.com/xebia-functional/xef/pull/196/commits/76d0e9ee0b87c8e10a8f7e2249397c00907cec5f)

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/chat/ChatCompletionRequestWithFunctions.kt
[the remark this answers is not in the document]
raulraja: @diesalbla going to resolve this conversation. If we want to attempt to change the model is fine, but we can do it in another PR. Feel free to take a shot at it if you think it can be improved.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
nomisRev: > That would be ideal, but based on our conversation on Slack yesterday I thought we still need the openAI deps in core if we want to provide a single module with sane defaults, and those defaults are the OpenAI ones.

I think there was some miscommunication there, `core` should expose some `interface` and `openAI` needs to expos some smart-constructor/implementation. Otherwise there is no point in splitting everything in different modules, if we end up enforcing depending on everything.

You wouldn't need to bring in two dependencies, since `OpenAI` would transitively bring in `core`, and unless versions start differing there is no need for `bom` IMO.

`getOrElse` is just there for backwards compatibility as requested by the other teams relying on Xef, it's nothing more than `try/catch` now so it can be removed entirely.

> Still, if we want xef-core to run independently without additional deps we need the OpenAI dep here or a different strategy.

Those decision are unrelated to technical side. Currently there is no clear vision of how we want this library to be used, or what the API should look like. We should answer that question first, and make the decisions in regards to that. If we want everyone to rely on `OpenAI` by default, or as a fallback, then it needs to be in `core`.

diesalbla:
> Note: I'm wondering if OpenAI-Kotlin implements Embeddings in a way that we can use in OpenAIEmbeddings.

There is indeed a class called `Embedding` in that package, namely `com.aallam.openai.api.embedding.Embedding`, and it has a cosine `similarity` extension method in the package `com.aallam.openai.client.extension`, so some can delegate there.

raulraja:
@nomisRev I adapted these changes to the latest from main after the Java stuff has been merged.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/AIRuntime.kt
```diff
@@ -1,31 +1,55 @@
...
+import com.xebia.functional.xef.llm.AIClient
+import com.xebia.functional.xef.llm.LLMModel
+import com.xebia.functional.xef.llm.openai.OpenAIClient
+import com.xebia.functional.xef.llm.openai.OpenAIEmbeddings
 import com.xebia.functional.xef.vectorstores.LocalVectorStore
 import kotlin.jvm.JvmStatic
 import kotlin.time.ExperimentalTime
 
-data class AIRuntime<A>(val runtime: suspend (block: AI<A>) -> A) {
+data class AIRuntime<A, C : AIClient>(
```
nomisRev, line 18:
```suggestion
data class AIRuntime<A>(
```

This generic parameter is unused, it's always used as `*`.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/AIClient.kt
```diff
@@ -0,0 +1,28 @@
...
+
+  suspend fun createChatCompletionWithFunctions(
+    request: ChatCompletionRequestWithFunctions
+  ): ChatCompletionResponseWithFunctions
+
+  suspend fun createEmbeddings(request: EmbeddingRequest): EmbeddingResult
+
+  suspend fun createImages(request: ImagesGenerationRequest): ImagesGenerationResponse
+
+  fun close(): Unit
```
nomisRev, line 27: Should this be exposed by Kotlin's `AutoCloseable`?

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/chat/Role.kt
```diff
@@ -0,0 +1,8 @@
+package com.xebia.functional.xef.llm.models.chat
+
+enum class Role {
+  system,
+  user,
+  assistant,
+  function
```
nomisRev, line 7:
```suggestion
  SYSTEM,
  USER,
  ASSISTANT,
  FUNCTION
```

We're using `ALLCAPS` and `AllCaps` notation, for `enum class`. It seems this is already inconsistent (we should fix that), the recommended style is `ALLCAPS` or `ALL_CAPS` (streaming snake case).

https://kotlinlang.org/docs/coding-conventions.html#property-names

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/embeddings/EmbeddingModel.kt
```diff
@@ -0,0 +1,5 @@
+package com.xebia.functional.xef.llm.models.embeddings
+
+enum class EmbeddingModel(val modelName: String) {
+  TextEmbeddingAda002("text-embedding-ada-002")
```
nomisRev, line 4:
```suggestion
  TEXT_EMBEDDING_ADA_002("text-embedding-ada-002")
```

Idem, this should use streaming snake case according to the official recommendations.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/OpenAIClient.kt
nomisRev: I'm not sure why that is necessary. _core_ should be independent of `open-ai`, right?

Since we have a domain to decouple _core_ from the `com.aallam.openai.api.BetaOpenAI` domain this implementation can live in a separate module that depends on _core_, right?

The other modules can then rely on this _only_ if they need the OpenAI dependency. `AIRuntime` also seems coupled to `xef-kotlin`, rather than _core_. In Java we need to think of a different strategy for dependency injection, perhaps Scala can re-use a similar idea but there we have `Using` in the Std.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/OpenAIEmbeddings.kt
```diff
@@ -22,7 +22,7 @@ class OpenAIEmbeddings(private val config: OpenAIConfig, private val oaiClient:
     }
     val lists: List<List<Embedding>> =
       if (texts.isEmpty()) emptyList()
-      else texts.chunked(chunkSize ?: config.chunkSize).parMap { createEmbeddings(it) }
+      else texts.chunked(chunkSize ?: 400).parMap { createEmbeddings(it) }
```
nomisRev, line 25: Why this introduction of a magic number, shouldn't the default exist inside of `OpenAIConfig` instead?

nomisRev (commented):
A couple of questions, and suggestions. Looks good overall. I am not sure I understand the dependency cycle between OpenAI and Core. It seems the dependencies need to be grouped in the top-level language specific modules if we want to provide top-level defaults rather than couple them in core.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/OpenAIEmbeddings.kt
raulraja: It does not look like it's available by the `OpenAIConfig` provided by the library.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/chat/Role.kt
raulraja: Adressed in https://github.com/xebia-functional/xef/pull/196/commits/ef1c24cac6bd635ff2fb8c08802fd75cdbf26697

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/embeddings/EmbeddingModel.kt
raulraja: Addressed in https://github.com/xebia-functional/xef/pull/196/commits/c5a70f0ecea5e0df60ce517109115a4b9a8ce956

raulraja:
> Besides a minor comment, the changes look good to me.
> 
> Note: I'm wondering if OpenAI-Kotlin implements `Embeddings` in a way that we can use in `OpenAIEmbeddings`.

This is how it is reflected now, the AIClient impl for OpenAI uses the library embeddings. Our embeddings delegate to the client embedding request, which delegates to the library.

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
```diff
@@ -41,31 +38,26 @@ public class AIScope implements AutoCloseable {
...
-        this(new ObjectMapper(), config, Executors.newCachedThreadPool(new AIScopeThreadFactory()));
+        this.client = runtime.getClient();
+        VectorStore vectorStore = new LocalVectorStore(runtime.getEmbeddings());
+        this.scope = new CoreAIScope(LLMModel.getGPT_3_5_TURBO(), LLMModel.getGPT_3_5_TURBO_FUNCTIONS(), client, vectorStore, runtime.getEmbeddings(), 3, "user", false, 0.4, 1, 20, 500);
     }
 
-    public AIScope(OpenAIConfig config, ExecutorService executorService) {
-        this(new ObjectMapper(), config, executorService);
+    public AIScope(AIRuntime runtime, ExecutorService executorService) {
+        this(new ObjectMapper(), runtime, Executors.newCachedThreadPool(new AIScopeThreadFactory()));
```
nomisRev, line 55: Overriding user provided `ExecutorService`.

```suggestion
    public AIScope(AIRuntime runtime, ExecutorService executorService) {
        this(new ObjectMapper(), runtime, executorService);
```

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
```diff
@@ -4,27 +4,24 @@
...
 import kotlin.collections.CollectionsKt;
 import kotlin.coroutines.Continuation;
 import kotlin.jvm.functions.Function1;
-import com.xebia.functional.xef.pdf.PDFLoaderKt;
-import kotlinx.coroutines.CoroutineScope;
-import kotlinx.coroutines.CoroutineStart;
-import kotlinx.coroutines.ExecutorsKt;
-import kotlinx.coroutines.JobKt;
-import kotlinx.coroutines.CoroutineScopeKt;
+import kotlinx.coroutines.*;
```
nomisRev, line 23: No start imports.

raulraja: Removed in https://github.com/xebia-functional/xef/pull/196/commits/f48bca33c96eaf6d389aa5e4371deb43366532bc

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/OpenAIEmbeddings.kt
raulraja: @diesalbla This is the code location we discussed regarding the magic number in today's meeting.

nomisRev (approved):

PR #197, opened 2023-06-22
Adding xef.java module

Yawolf:
The idea is to have this in Java:
```java
class Person {
  public String name;
  public Int age;
}

try(AIScope scope = AIScope(...)) {
  Person person = scope.prompt("Give me a person", Person.class, ...)
  ...
}
```

`AIScope` from `xef-core` has been renamed to `CoreAIScope` for avoiding name collisions in Java and because it's a piece, from core project, needed for creating/instantiating specific AIScope in other implementations. `BaseAIScope` could work too.

`CoreAIScope` is now an `AutoCloseable` object so it can be used with a `try-with-resources` form Java.

- [x] Adding a new module for Java implementation
- [x] `prompt` compiles and is working
- [ ] `promptMessage` compiles and is working
- [ ] `image` compiles and is working
- [ ] `pdf` compiles and is working
- [ ] `contextScope` compiles and is working


nomisRev:
This PR is now ready for review, and ready to be merged. We want to prevent this PR from growing too large, and we need to investigate what we want to do with PDF, etc in terms of architecture.

When this is merged we can start porting more Java examples.

diesalbla (commented):
Couple suggestions...

raulraja (dismissed):
Looks great @nomisRev @Yawolf !

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private <T> T undefined() {
+        throw new RuntimeException("Method is undefined");
+    }
+
+    public <A> A prompt(String prompt, Class<A> cls) {
+        return prompt(prompt, cls, coreAIScope.getMaxDeserializationAttempts(), coreAIScope.getDefaultSerializationModel(), coreAIScope.getUser(), coreAIScope.getEcho(), coreAIScope.getNumberOfPredictions(), coreAIScope.getTemperature(), coreAIScope.getDocsInContext(), coreAIScope.getMinResponseTokens());
+    }
+
+    public <A> A prompt(String prompt, Class<A> cls, Integer maxAttempts, LLMModel llmModel, String user, Boolean echo, Integer n, Double temperature, Integer bringFromContext, Integer minResponseTokens) {
```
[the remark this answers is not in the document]
Yawolf: I didn't know that `record` was a valid keyword in Java. I can use it in further examples rather than using public classes with public static fields 😄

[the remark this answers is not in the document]
diesalbla: Word of caution, though, is that this feature is only available in versions after Java 16, so perhaps using it would restrict which JVM versions can use this library.

Is there a decision as to what versions of the JVM would this be targetting?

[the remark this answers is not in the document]
nomisRev: @diesalbla, good point. We were planning to remove the Loom requirement as well, but perhaps we can just choose a higher JDK version instead. JDK 21 is around the corner, but perhaps a bit too ambitious. Having LOOM would be quite neat though, could get some _async_ hurdles out of the way.

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
```diff
@@ -0,0 +1,147 @@
...
+        OpenAIConfig config = new OpenAIConfig();
+        KtorOpenAIClient client = new KtorOpenAIClient(config);
+        try {
+            Embeddings embeddings = new OpenAIEmbeddings(config, client);
+            VectorStore vectorStore = new LocalVectorStore(embeddings);
+            CoreAIScope scope = new CoreAIScope(LLMModel.getGPT_3_5_TURBO(), LLMModel.getGPT_3_5_TURBO_FUNCTIONS(), client, vectorStore, embeddings, 3, "user", false, 0.4, 1, 20, 500);
+            return block.invoke(new AIScope(scope));
+        } finally {
+            client.close();
+        }
```
Yawolf, line 56: Maybe we could add here some `try-with-resources` syntax?
```java
try (KtorOpenAIClient client = new KtorOpenAIClient(config)) {
  // the code
}
```

nomisRev: Yes, that was my original intention but I need to investigate what is going on under the hood. Java doesn't recognise `KtorOpenAIClient` as implementing Java's `AutoCloseable` but that might've just been IDEA acting up.

Yawolf: I've been reading and it seems like, although `AutoCloseable` in Kotlin and Java behave in the same way, they're not the same interface, maybe that's the reason Java didn't recognise `KtorOpenAIClient` as a `AutoCloseable`

nomisRev: Aha, it seems you're right but this also seems like a temporary measure 🤔 Normally it's possible for it to be the same type, but it seems it's not the case. I made some updates to the PR that resolves this issue though and you can now use `AIScope` with `try-with-resources` syntax.

https://github.com/JetBrains/kotlin/blob/a9d0ff83cf7dc847213c31b61b7d0c88ab62a95c/libraries/stdlib/jdk7/src/kotlin/AutoCloseableJVM.kt#L20

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/models.kt
```diff
@@ -89,6 +97,52 @@ data class ChatCompletionRequestWithFunctions(
...
+            "enum": ["celsius", "fahrenheit"]
+          }
+        },
+        "required": ["location"]
+      }
+    }
+  ]
+ */
+@Serializable
+data class CFunction(
```
diesalbla, line 123: This seems to be duplicated from the file `com.xebia.functional.xef.llm.openai.functions.CFunction`. What is the relation between them?

diesalbla (dismissed):

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/openai/models.kt
nomisRev: That was a mistake during merging, I removed the duplicated model in `com.xebia.functional.xef.llm.openai.functions.CFunction`. Thank you for catching this @diesalbla!

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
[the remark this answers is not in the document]
nomisRev: See discussion on Slack, we decided on JDK8. https://fortyseven.slack.com/archives/C051PG86AH0/p1687795788974039?thread_ts=1687789966.900839&cid=C051PG86AH0

diesalbla (approved):

PR #198, opened 2023-06-22
TokenTextSplitter Ref: extract method, use range.

diesalbla:


core/src/commonMain/kotlin/com/xebia/functional/xef/textsplitters/TokenTextSplitter.kt
```diff
@@ -14,14 +15,12 @@ private class TokenTextSplitterImpl(
...
-      .asSequence()
-      .filter { it % stepSize == 0 }
-      .map { startIdx -> inputIds.subList(startIdx, minOf(startIdx + chunkSize, inputIds.size)) }
-      .map { chunkIds -> tokenizer.decode(chunkIds) }
-      .toList()
+    fun decodeSegment(startIdx: Int): String {
+      val end = minOf(startIdx + chunkSize, inputIds.size)
+      return tokenizer.decode(inputIds.subList(startIdx, end))
+    }
+    val segments = IntProgression.fromClosedRange(0, inputIds.size - 1, chunkSize - chunkOverlap)
```
nomisRev, line 22: Nit: at the cost of 1 additional allocation, we get slightly better readability? 🤔
 
```suggestion
    val segments = inputIds.indices step (chunkSize - chunkOverlap)
```

core/src/commonMain/kotlin/com/xebia/functional/xef/textsplitters/TokenTextSplitter.kt
```diff
@@ -14,14 +15,12 @@ private class TokenTextSplitterImpl(
...
+    return segments.toList().map { decodeSegment(it) }
```
nomisRev, line 23: Since `Range` is an `Iterable` we have `map` directly available here which returns a `List` after its operation. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map.html

```suggestion
    return segments.map { decodeSegment(it) }
```

nomisRev (approved):
Nice refactor. Just 1 nit, and a suggestion ☺️
Thanks @diesalbla

nomisRev (approved):

PR #199, opened 2023-06-22
OpenAIEmbeddings - Refactor

diesalbla:
Nest logic internally. Use nested functions.

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/embeddings/OpenAIEmbeddings.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/embeddings/OpenAIEmbeddings.kt
@@ -15,27 +15,17 @@ class OpenAIEmbeddings(private val config: OpenAIConfig, private val oaiClient: 
     texts: List<String>,
     chunkSize: Int?,
     requestConfig: RequestConfig
-  ): List<Embedding> = chunkedEmbedDocuments(texts, chunkSize ?: config.chunkSize, requestConfig)
+  ): List<Embedding> {
+    suspend fun createEmbeddings(texts: List<String>): List<Embedding> {
+      val req = EmbeddingRequest(requestConfig.model.modelName, texts, requestConfig.user.id)
+      return oaiClient.createEmbeddings(req).data.map { Embedding(it.embedding) }
+    }
+    val lists: List<List<Embedding>> =
+      if (texts.isEmpty()) emptyList()
+      else texts.chunked(chunkSize ?: config.chunkSize).parMap { createEmbeddings(it) }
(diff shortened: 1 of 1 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/embeddings/OpenAIEmbeddings.kt
```diff
@@ -15,27 +15,19 @@ class OpenAIEmbeddings(private val config: OpenAIConfig, private val oaiClient:
     texts: List<String>,
     chunkSize: Int?,
     requestConfig: RequestConfig
-  ): List<Embedding> = chunkedEmbedDocuments(texts, chunkSize ?: config.chunkSize, requestConfig)
+  ): List<Embedding> {
+    inline fun toRequest(texts: List<String>): EmbeddingRequest =
```
nomisRev, line 19: > Local inline functions are not yet supported in inline functions

nomisRev, line 19: > Local inline functions are not yet supported in inline functions

```suggestion
   fun toRequest(texts: List<String>): EmbeddingRequest =
```

core/src/commonMain/kotlin/com/xebia/functional/xef/embeddings/OpenAIEmbeddings.kt
```diff
@@ -15,27 +15,19 @@ class OpenAIEmbeddings(private val config: OpenAIConfig, private val oaiClient:
...
+      EmbeddingRequest(requestConfig.model.modelName, texts, requestConfig.user.id)
 
-  override suspend fun embedQuery(text: String, requestConfig: RequestConfig): List<Embedding> =
-    if (text.isNotEmpty()) embedDocuments(listOf(text), null, requestConfig) else emptyList()
+    suspend fun createWithRetry(texts: List<String>): List<Embedding> =
```
nomisRev, line 22: Nit: This `withRetry` is a bit outdated since that logic is no longer present. It exists directly inside the `HttpClient` now. Perhaps just `createEmbeddings`.

This suggest breaks the code below.

```suggestion
    suspend fun createEmbeddings(texts: List<String>): List<Embedding> =
```

core/src/commonMain/kotlin/com/xebia/functional/xef/embeddings/OpenAIEmbeddings.kt
diesalbla: Rather strange limitation... ok.

PR #206, opened 2023-06-27
Rename Core's AIScope to CoreAIScope

nomisRev:
This PR renames `AIScope` to `CoreAIScope` from `xef-core`. The goal is to have a distinct name for the `AIScope` defined in `xef-core`, such that all languages can expose `AIScope` in the same package if they want.

On Scala, and Java composition over inheritance will be preferred to expose Scala and Java APIs. In Kotlin we can simply leverage extension functions, and a `typealias AIScope`.


diesalbla (approved):

diesalbla:
Thanks!

PR #213, opened 2023-06-30
Delegate Cosine Similarity to OpenAI-Kotlin

diesalbla:
The OpenAI-Kotlin library https://github.com/aallam/openai-kotlin implements the cosine similarity feature for "embeddings", so we could delegate to it. However the API is experimental and we would need to couple to it in the `core`.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -64,10 +64,9 @@ kotlin {
     val commonMain by getting {
       dependencies {
         api(libs.bundles.arrow)
-
         api(libs.bundles.ktor.client)
         api(projects.xefTokenizer)
-
+        implementation(libs.openai.client)
         implementation(libs.uuid)
         implementation(libs.klogging)
       }
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
@@ -2,11 +2,13 @@ package com.xebia.functional.xef.vectorstores
 
 import arrow.atomic.Atomic
 import arrow.atomic.getAndUpdate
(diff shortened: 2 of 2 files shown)


openai/src/commonMain/kotlin/com/xebia/functional/xef/auto/llm/openai/OpenAIRuntime.kt
```diff
@@ -19,15 +19,7 @@ object OpenAIRuntime {
   @OptIn(ExperimentalTime::class)
   @JvmStatic
   fun <A> openAI(config: OpenAIConfig? = null): AIRuntime<A> {
```
nomisRev, line 21:
```suggestion
  fun <A> openAI(config: OpenAIConfig = defaultConfig()): AIRuntime<A> {
```

raulraja:
I think this PR is still relevant @diesalbla , ping me for review when you update it to the latest from `main`. thanks!

diesalbla:
This may be ready for review now.

core/build.gradle.kts
```diff
@@ -64,10 +64,9 @@ kotlin {
     val commonMain by getting {
       dependencies {
         api(libs.bundles.arrow)
-
         api(libs.bundles.ktor.client)
         api(projects.xefTokenizer)
-
+        implementation(libs.openai.client)
```
raulraja, line 69: This couples core with open ai, which we tried to avoid and worked to decouple in the last weeks.
Looking at what is delegating to https://github.com/aallam/openai-kotlin/blob/main/openai-client/src/commonMain/kotlin/com.aallam.openai.client/extension/internal/CosineSimilarity.kt#L19 it seems simple enough not to include as a dependency. We could provide similar functions locally or change ours to that implementation if it's better than what we have now.

serras (changes requested):
I am pretty unsure of this PR, cosine similarity is a function simple enough that we should not depend on any library.

Issue #215, opened 2023-07-02
Keep ggml and llama.cpp dependencies up to date.


raulraja:
Currently, the llama.cpp binaries are stored in the resources folder for what looks now an old version.

I noticed that when trying to load many of the ggml model files from huggingface many of them fail. We don't know if is due to these binaries being outdated. Ideally, we can point to a release version we compile automatically in CI or we fetch the binaries if they are published somewhere.

<img width="292" alt="Screenshot 2023-07-02 at 20 03 02" src="https://github.com/xebia-functional/xef/assets/456796/321532f3-ba26-414c-8e2f-8b73bc2c0178">

For example:
```kotlin
package com.xebia.functional.xef.auto.gpt4all

import com.xebia.functional.gpt4all.GPT4All
import com.xebia.functional.gpt4all.LLModel
import com.xebia.functional.gpt4all.getOrThrow
import com.xebia.functional.gpt4all.huggingFaceUrl
import com.xebia.functional.xef.auto.PromptConfiguration
import com.xebia.functional.xef.auto.ai
import com.xebia.functional.xef.auto.llm.openai.OpenAI
import com.xebia.functional.xef.pdf.pdf
import java.nio.file.Path

suspend fun main() {
  //TheBloke/mpt-30B-chat-GGML
  val userDir = System.getProperty("user.dir")
  val path = "$userDir/models/gpt4all/Wizard-Vicuna-13B-Uncensored.ggmlv3.q8_0.bin"
  val url = "https://huggingface.co/TheBloke/Wizard-Vicuna-13B-Uncensored-GGML/resolve/main/Wizard-Vicuna-13B-Uncensored.ggmlv3.q8_0.bin"
  val modelType = LLModel.Type.GPTJ
  val modelPath: Path = Path.of(path)
  val GPT4All = GPT4All(url, modelPath, modelType)

  println("🤖 GPT4All loaded: $GPT4All")

  val pdfUrl = "https://kotlinlang.org/docs/kotlin-reference.pdf"

  /**
   * Uses internally [HuggingFaceLocalEmbeddings] default of "sentence-transformers", "msmarco-distilbert-dot-v5"
   * to provide embeddings for docs in contextScope.
   */

  ai {
    println("🤖 Loading PDF: $pdfUrl")
    contextScope(pdf(pdfUrl)) {
      println("🤖 Context loaded: $context")
      GPT4All.use { gpT4All: GPT4All ->
        println("🤖 Generating prompt for context")
[...]

PR #216, opened 2023-07-03
GPT4All Java Bindings and supported models list

raulraja:
Fixes https://github.com/xebia-functional/xef/issues/215

This PR introduces the official GPT4ALL Java bindings, which provide access to GPT4All functionality. With the introduction of these bindings, we no longer need to maintain the `llama.cpp` binaries. The changes in this PR remove references to `llama.cpp` and replace them with the new GPT4ALL Java bindings.

The key changes in this PR are as follows:

- Added the `Gpt4AllModel` class, which represents a GPT4All model and provides methods to fetch supported models from a JSON file.
- Removed the `GPT4AllModel` interface, as it is no longer needed.
- Removed the `LLModel` interface and its implementations (`LlamaModel` and `GPTJModel`), as they are no longer used.
- Updated the `GPT4All` interface to use the new GPT4ALL Java bindings.
- Modified the `GPT4All.invoke()` function to load the GPT4All model using the new bindings.
- Updated the `createCompletion()` and `createChatCompletion()` functions in the `GPT4All` interface to use the new GPT4ALL Java bindings for generating completions.
- Updated the `main()` function in the `Chat.kt` example to use the new GPT4ALL Java bindings and the `Gpt4AllModel.supportedModels()` function to display [the supported models](https://github.com/nomic-ai/gpt4all/blob/main/gpt4all-chat/metadata/models.json).
- Removed references to `llama.cpp` and related binaries from the codebase.
- Updated the build configuration to include the Kotlin Serialization plugin.


raulraja:
@xebia-functional/team-ai

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/Models.kt
```diff
@@ -0,0 +1,32 @@
...
+  val parameters: String,
+  val quant: String,
+  val type: String,
+  val description: String,
+  val disableGUI: String? = null,
+  val url: String? = null
+) {
+  companion object {
+    private val url = "https://raw.githubusercontent.com/nomic-ai/gpt4all/main/gpt4all-chat/metadata/models.json"
+    fun supportedModels(): List<Gpt4AllModel> {
```
diesalbla, line 25: Is this going to run the HTTP operation and fetch the list of supported models from that `URL` on every call?

raulraja: For now, yes, We have discussed the concept of a Model Zoo / List that we support beyond the GPT4ALL ones with our curated list of models. Feel free to commit any improvements directly to this branch if we want to cache the results. I may not have much more time today to dedicate to it.

diesalbla (approved):

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/Models.kt
raulraja: Let's address this when we have a strategy for our recommended models.

PR #224, opened 2023-07-04
Remove publish plugin for xef-kotlin module

franciscodr:
The `xef-kotlin` module is not ready yet, preventing other modules from publishing in Maven. The reason is that the task for generating the metadata file fails as the module doesn't contain code.

This pull request disables the publish plugin temporarily.

### DIFF
--- a/kotlin/build.gradle.kts
+++ b/kotlin/build.gradle.kts
@@ -11,7 +11,6 @@ plugins {
   alias(libs.plugins.kotlinx.serialization)
   alias(libs.plugins.spotless)
   alias(libs.plugins.dokka)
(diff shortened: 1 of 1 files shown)


raulraja (approved):
thanks @franciscodr for finding the cause 🙏

diesalbla (approved):

raulraja:
Please do not merge this, as I have moved stuff to that module in #225 which addresses this issue

raulraja (changes requested):
See #225

raulraja:
This was addressed in #225 and can be closed if https://github.com/xebia-functional/xef/actions/runs/5453261457 publishes a new alpha

PR #225, opened 2023-07-04
Move `ai` block to xef-kotlin module

raulraja:
This PR includes the following changes and removes the need to merge #224 

- In the file `AI.kt`:
  - Removed the `AI` type alias declaration.
  - Removed the inline function `ai`.

- In the file `Gpt4AllRuntime.kt`:
  - Removed unused imports.
  - Updated the implementation of the `toEither` function to use the `Either.catchOrThrow` operator instead of the `ai` function.
  - Removed the unused import and the `ai` function.

- Added a new file `DSLExtensions.kt` to the `kotlin/src/commonMain/kotlin/com/xebia/functional/xef/auto` directory, which contains:
  - A new inline function `ai` that provides a DSL block for constructing `AI` values.

- In the file `OpenAIRuntime.kt`:
  - Removed unused imports.
  - Updated the implementation of the `toEither` function to use the `Either.catchOrThrow` operator instead of the `ai` function.
  - Removed the `@OptIn(ExperimentalTime::class)` annotation.

### DIFF
--- a/gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/Gpt4AllRuntime.kt
+++ b/gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/Gpt4AllRuntime.kt
@@ -1,12 +1,9 @@
 package com.xebia.functional.gpt4all
 
 import arrow.core.Either
-import arrow.core.left
-import arrow.core.right
 import com.xebia.functional.xef.AIError
 import com.xebia.functional.xef.auto.AI
 import com.xebia.functional.xef.auto.CoreAIScope
-import com.xebia.functional.xef.auto.ai
 
 /**
  * Run the [AI] value to produce an [A], this method initialises all the dependencies required to
@@ -38,7 +35,7 @@ suspend inline fun <reified A> AI<A>.getOrThrow(): A = getOrElse { throw it }
  * @see getOrElse for an operator that allow directly handling the [AIError] case.
  */
 suspend inline fun <reified A> AI<A>.toEither(): Either<AIError, A> =
-  ai { invoke().right() }.getOrElse { it.left() }
+  Either.catchOrThrow { getOrThrow() }
 
 suspend fun <A> AIScope(block: AI<A>, orElse: suspend (AIError) -> A): A =
   try {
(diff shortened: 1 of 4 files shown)


gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/Gpt4AllRuntime.kt
```diff
@@ -38,7 +35,7 @@ suspend inline fun <reified A> AI<A>.getOrThrow(): A = getOrElse { throw it }
  * @see getOrElse for an operator that allow directly handling the [AIError] case.
  */
 suspend inline fun <reified A> AI<A>.toEither(): Either<AIError, A> =
```
nomisRev, line 37: I don't think this `A` needs to be `reified`? 🤔

openai/src/commonMain/kotlin/com/xebia/functional/xef/auto/llm/openai/OpenAIRuntime.kt
```diff
@@ -42,9 +38,8 @@ suspend inline fun <reified A> AI<A>.getOrThrow(): A = getOrElse { throw it }
  * @see getOrElse for an operator that allow directly handling the [AIError] case.
  */
 suspend inline fun <reified A> AI<A>.toEither(): Either<AIError, A> =
```
nomisRev, line 40: Idem: `A` shouldn't be reified.

nomisRev (approved):
Looks good! I think the `AI<A>` alias can also be moved.

Two unrelated comments to this PR but my eye fel on it. Can be fixed here, or in a subsequent PR.

Yawolf (approved):
Once Simon's suggestion are done 🙂

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/Gpt4AllRuntime.kt
raulraja: Probably it doesn't and comes from the original impl, where it was serializing automatically all over.

openai/src/commonMain/kotlin/com/xebia/functional/xef/auto/llm/openai/OpenAIRuntime.kt
raulraja: Makes sense. Let's address this on a different PR since they are unrelated.

PR #226, opened 2023-07-04
`promptMessage` alternative that just returns `String`

raulraja:


openai/src/commonMain/kotlin/com/xebia/functional/xef/auto/llm/openai/OpenAIScopeExtensions.kt
```diff
@@ -9,6 +9,13 @@ import com.xebia.functional.xef.llm.models.functions.CFunction
 import com.xebia.functional.xef.prompt.Prompt
 import kotlinx.serialization.serializer
 
+@AiDsl
+suspend fun CoreAIScope.promptMessage(
+  prompt: String,
+  model: Chat = OpenAI.DEFAULT_CHAT,
+  promptConfiguration: PromptConfiguration = PromptConfiguration.DEFAULTS,
+): String = model.promptMessage(prompt, context, promptConfiguration)
+
```
Yawolf, line 18: Question! Why don't you better include this into the `CoreAIScope` as an Instance Method rather than an Extension Method?

raulraja: This is already there as well, and here it appears as a convenience to the user where it does not need to specify models and uses the defaults:
```kotlin
package com.xebia.functional.xef.auto

import com.xebia.functional.xef.auto.llm.openai.getOrElse
import com.xebia.functional.xef.auto.llm.openai.promptMessage

suspend fun main() {
  ai {
    val love: String = promptMessage("tell me you like me with just emojis")
    println(love)
  }.getOrElse { println(it) }
}
```
If you don't use the open ai imports then you have to do:
```kotlin
package com.xebia.functional.xef.auto

import com.xebia.functional.xef.auto.llm.openai.getOrElse

suspend fun main() {
  ai {
    val love: String = OpenAI.DEFAULT_CHAT.promptMessage("tell me you like me with just emojis")
    println(love)
  }.getOrElse { println(it) }
}
```

Yawolf (approved):

Yawolf (dismissed):

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
```diff
@@ -17,6 +17,15 @@ interface Chat : LLM {
 
   fun tokensFromMessages(messages: List<Message>): Int
 
+  @AiDsl
+  suspend fun promptMessage(
```
serras, line 21: Isn't this conflicting with the function of the same name returning a `List`? My experience is that overloading by return type usually leads to problems, maybe we can think of another name?

raulraja: I found no conflicts, probably because all the use sites had type descriptions. 
Just in case, to address this I changed in [22c8a1e](https://github.com/xebia-functional/xef/pull/226/commits/22c8a1e8138a070724cd3d2e84252c6727e36a87) `promptMessage` to return a `String` and `promptMesages` to return a `List<String>` so there are no conflicts.

raulraja:
Going to merge this as it includes a fix for the GPT4ALL supported model loading. Let me know if you want me to address anything else and I'll open a new PR.

PR #234, opened 2023-07-11
Improving Json Schema generation, and definition in Java

nomisRev:
This PR improves the way we define, and generate Json Schema from Json. The Jackson-schema library doesn't support V7 schema which is required for OpenAI.

The new library supports V7, and also allows to easily define `required` fields using Jakarta `NonNull` annotation. It furthermore also allows defining different patterns which will also be included in the Json schema send to OpenAPI.

This is needed because otherwise in a lot of cases we saw `null` values in the responses from the AI, which was not happening in Scala or Kotlin thanks to gullible types and Option type.


examples/java/src/main/java/com/xebia/functional/xef/java/auto/Books.java
```diff
@@ -0,0 +1,45 @@
...
+                    '}';
+        }
+    }
+
+    public CompletableFuture<Books.Book> bookSelection(String topic) {
+        return scope.prompt("Give me a selection of books about " + topic, Books.Book.class);
+    }
+
+    public static void main(String[] args) throws ExecutionException, InterruptedException {
+        try (AIScope scope = new AIScope()) {
```
Zevleg, line 38: If you are not going to use scope, why do you create the object? you are not using it hehe

PR #253, opened 2023-07-17
Java example gpt4all

Zevleg:
I haven't could implement the use of Flow in Java, it seems that Java doesn't support .onCompletion and you have to send a FlowCollector and a Continuation to  use .collect(), so because of that I implemented a kotlinPort to use Flow in kotlin and getting the answer in Java.

Ticket [Java Example (GPT4ALL)](https://app.clickup.com/t/861n05vep)


examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
```diff
@@ -0,0 +1,58 @@
...
+
+        var supportedModels = Gpt4AllModel.Companion.getSupportedModels();
+
+        supportedModels.forEach(it -> {
+            var url = (Objects.nonNull(it.getUrl())) ? " - " + it.getUrl() : "";
+            System.out.println("🤖 " + it.getName() + url);
+        });
+
+        var url = "https://huggingface.co/nomic-ai/ggml-replit-code-v1-3b/resolve/main/ggml-replit-code-v1-3b.bin";
+        var modelPath= Path.of(path);
```
victorcrrd, line 28:
```suggestion
        var modelPath = Path.of(path);
```

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
```diff
@@ -0,0 +1,58 @@
...
+         * to provide embeddings for docs in contextScope.
+         */
+
+        try (AIScope scope = new AIScope()) {
+            System.out.println("🤖 Context loaded: " + scope.getContext());
+
+            while(true){
+                System.out.println("\n🤖 Enter your question: ");
+
+                BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
```
victorcrrd, line 43: I'm not really sure about this, but can we move the declaration of the `BufferedReader` to outside the `while` loop? Isn't this creating a new instance for each line to be read?

victorcrrd (commented):
Some comments. I don't know about the use of `KotlinPort`, so I will let others review that :)

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
```diff
@@ -0,0 +1,58 @@
...
+                String line = br.readLine();
+                if (line == null || line.isBlank()) {
+                    break;
+                }else{
+                    var promptConfiguration = PromptConfiguration.Companion.buildWithParams(2, true);
```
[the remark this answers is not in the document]
Zevleg: Done

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/port/KotlinPort.kt
```diff
@@ -0,0 +1,30 @@
...
+    val answer: MutableList<String> = mutableListOf()
+    promptStreaming.onCompletion {
+        println("\n🤖 Done")
+    }.collect {
+        answer.add(it)
+        System.out.print(it)
+    }
+
+    return answer
+}
```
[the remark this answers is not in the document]
Zevleg: Done, just to know, why?

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
```diff
@@ -121,14 +127,17 @@ public CompletableFuture<List<String>> promptMessages(Chat llmModel, String prom
         return future(continuation -> scope.promptMessages(llmModel, prompt, functions, promptConfiguration, continuation));
     }
 
+    public CompletableFuture<List<String>> promptStreaming(GPT4All gpt4all, String line, VectorStore context, PromptConfiguration promptConfiguration) {
```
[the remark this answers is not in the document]
Zevleg: it's not necessary, Done

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
Zevleg: Done

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
Zevleg: Done

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/port/KotlinPort.kt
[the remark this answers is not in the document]
victorcrrd: Somehow I got that habit. But I remember doing `cat sourceCodeFile` to files without that ending line and the terminal doesn't include a new line in the end, so new commands are in that same last line 😅

examples/java/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ plugins {
 
 dependencies {
     implementation(projects.xefJava)
+    implementation(libs.kotlinx.coroutines.reactive)
```
raulraja, line 10: Since the `promptStreaming : Flow<String>` support is available in xef-core this dependency should appear as part of the `xef-java` example instead of the examples. `xef-java` is the one providing `promptStreaming: Publisher<String>`

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
```diff
@@ -0,0 +1,87 @@
...
+        System.out.println("🤖 GPT4All loaded: " + gpt4all);
+        /**
+         * Uses internally [HuggingFaceLocalEmbeddings] default of "sentence-transformers", "msmarco-distilbert-dot-v5"
+         * to provide embeddings for docs in contextScope.
+         */
+
+        try (AIScope scope = new AIScope()) {
+            System.out.println("🤖 Context loaded: " + scope.getContext());
+
+            BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
```
raulraja, line 42: Both `BufferedReader` and `InputStreamReader` implement `AutoCloseable` and should be specified in `try-with-resources` blocks so they can be properly finalized and closed when done.

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
```diff
@@ -0,0 +1,87 @@
...
+
+//                    answer.forEach(it -> {
+//                        System.out.print(it);
+//                    });
+
+                    answer.subscribe(new Subscriber<String>() {
+                        StringBuilder answer = new StringBuilder();
+                        @Override
+                        public void onSubscribe(Subscription s) {
+                            System.out.println("\n🤖 --> " + s);
```
raulraja, line 62: instead of using one line per character or word getting streamed back, we should use here ` System.out.print(s);` to avoid new lines after each streamed chunked. This could be better observed if you use the OpenAI models in the example instead of GPT4All, as the streaming there is more reliable.

gpt4all-kotlin/build.gradle.kts
```diff
@@ -64,6 +64,7 @@ kotlin {
       dependencies {
         implementation(libs.gpt4all.java.bindings)
         implementation(libs.ai.djl.huggingface.tokenizers)
+        implementation(libs.kotlinx.coroutines.reactive)
```
raulraja, line 67: This should be part of `xef-java`, not here. We don't want the reactive streams `Publisher` dependency anywhere else but where it makes sense, which is the Java integration. Other integrations like Kotlin or Scala do not need or use `Publisher` or anything from the Reactive Streams spec.

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/port/KotlinPort.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.xebia.functional.gpt4all.GPT4All
+import com.xebia.functional.xef.auto.AiDsl
+import com.xebia.functional.xef.auto.PromptConfiguration
+import com.xebia.functional.xef.prompt.Prompt
+import com.xebia.functional.xef.vectorstores.VectorStore
+import kotlinx.coroutines.reactive.asPublisher
+import org.reactivestreams.Publisher
+
+@AiDsl
+suspend fun promptStreaming(
```
raulraja, line 14: Should be in `xef-java` as operator similar how `promptMessages` and `prompt` are declared now.

java/build.gradle.kts
```diff
@@ -13,10 +13,12 @@ dependencies {
     api(projects.xefCore)
     api(projects.xefOpenai)
     api(projects.xefPdf)
+    api(projects.xefGpt4all)
```
raulraja, line 16: xef-java should not depend on the GPT4All integration. I need to see a table or schema on how you are all planning these integrations because currently the dependencies associations in this PR do not match the plans on making Java as lean as possible and not depend on another module that is not `xef-core` and Java specifics deps.

A module like `xef-java` should not depend on the rest of the modules. If another module like `xef-gpt4all` written in Kotlin is not properly encoded to be consumed from Java then it should be fixed there instead of bringing it as a dependency. The idea is that users of `xef-kotlin`, `xef-java` or `xef-scala` can import `xef-gpt4all` as dependency and use it without rewriting its apis. We don't want to port every API we develop in modules like `xef-gpt4All` or `xef-openai` to every language but instead, use them where possible. 

In the case of Java, Kotlin multiplatform has all the tools to export the types through @JvmName, @JvmStatic and @JvmFields as things that can be easily consumed from Java and other JVM langs like Scala. Only functions from `xef-core` should be ported to Java where needed if they use continuations or `suspend` for Kotlin. // cc @franciscodr @nomisRev for thoughts on how this should work.

raulraja (changes requested):
I left some comments regarding dependencies we need to agree on before merging this.
We need to all get on the same page about what role each module serves and how dependencies between modules should be organized.

In my understanding, the goals when it comes to implementing something like this are:
1. `xef-core` has no dependencies and is the base building blocks for other langs
2. Integrations are developed in Kotlin multiplatform. For example `xef-openai`, `xef-gpt4all` with proper annotations support like @JvmInline etc.
3. Functions like `promptStreaming: Flow<String>` currently declared in core if they were to be ported over to Java in that case the adapter `promptStreaming: Publisher<String>` lives in the `xef-java` module.
4. Users that want to use something like `gpt4all` in Java all they have to do is bring as dependencies `xef-java` and `xef-gpt4all`. Since `promptStreaming: Pubisher<String>` is declared in `xef-java` and `xef-gpt4all` provides `Chat` model impl then they can just use the function passing in the model or in whichever this function is exposed in `xef-java`

java/build.gradle.kts
nomisRev: I agree, but not sure it will fit all use-cases. I.e. `Flow` is another one that might need wrapping and `@JvmName` or other Java annotations won't get use far enough.

We can of course expose Java specific implementation from `jvmMain`, but the biggest downside is that they will appear as extension functions and not class methods. Wrapping gives us a better and nicer APIs in those areas. I'm kind-of missing [C# partial classes](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods) for KMP implementations. Could make an interesting KEEP.

Scala has a similar issue where we might want to expose FS2 instead of `Flow`, or a different streaming data type. So we might end up wrapping some of the APIs, and exposing them from different modules to adapt the return types. However the underlying implementation would still be common.

Zevleg: if I remove api(projects.xefGpt4all) from xef-java I can not import GPT4All and Gpt4AllModel and they are needed

Zevleg: Also I could impl a promptStreaming in Chat.kt from xef-core but I would have to add the Publisher dependency, or add the implementation in CoreAIScope.kt but I would have to add the GPT4All dependency

@AiDsl
  suspend fun promptStreaming(
    question: String,
    context: VectorStore,
  ): Publisher<String> =
    promptStreaming(Prompt(question), context, null, emptyList(), promptConfiguration).asPublisher()

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/port/KotlinPort.kt
Zevleg: I tried to implement in CoreAIScope like promptMessages but I have to add the GPT4ALL dependency in xef-core

```kotlin
@AiDsl
suspend fun promptStreaming(
    gpt4all: GPT4All,
    question: String,
    context: VectorStore,
): Publisher<String> = gpt4all.promptStreaming(Prompt(question), context, null, emptyList(), promptConfiguration).asPublisher()
```

Zevleg: also I move KotlinPort.kt to xef-java but I'm having this error: 

```java
/java/auto/AIScope.java:15: error: package com.xebia.functional.xef.java.auto.port does not exist
import com.xebia.functional.xef.java.auto.port.KotlinPort;
```

examples/java/build.gradle.kts
Zevleg: Done

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
Zevleg: Done

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
Zevleg: Done

gpt4all-kotlin/build.gradle.kts
Zevleg: Done

java/build.gradle.kts
raulraja: `projects.xefGpt4all` should only be added as a dependency in the examples.
xef-java should add the dependency for Reactive streams and implement `promptStreaming` in xef-java.

Zevleg: Done

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/port/KotlinPort.kt
Zevleg: Implemented how promptMessages, Done

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
```diff
@@ -121,14 +128,17 @@ public CompletableFuture<List<String>> promptMessages(Chat llmModel, String prom
         return future(continuation -> scope.promptMessages(llmModel, prompt, functions, promptConfiguration, continuation));
     }
 
+    public CompletableFuture<Publisher<String>> promptStreaming(GPT4All gpt4all, String line, PromptConfiguration promptConfiguration) {
```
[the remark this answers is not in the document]
Zevleg: now is returning CompletableFuture<Flow<String>> and in the class Chat.java it's implementing:
```java
ReactiveFlowKt.asPublisher(answer)
```

examples/java/src/main/java/com/xebia/functional/xef/java/auto/gpt4all/Chat.java
```diff
@@ -0,0 +1,83 @@
...
+            while(true){
+                System.out.println("\n🤖 Enter your question: ");
+
+                String line = br.readLine();
+                if (line.equals("exit")) break;
+
+                var promptConfiguration = new PromptConfiguration.Companion.Builder().docsInContext(2).streamToStandardOut(true).build();
+                Flow<String> answer = scope.promptStreaming(gpt4all, line, promptConfiguration).get();
+
+                ReactiveFlowKt.asPublisher(answer).subscribe(new Subscriber<String>() {
```
raulraja, line 55: This part should not go on the example but in the `xef-java` project. where you implement `promptStreaming: Publisher<String>`. Whoever uses this API from Java is not exposed directly to the `kotlin` Flow type but to the Java ones.
Whoever uses the examples already has this API that returns the Publisher, and they add a subscription.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
```diff
@@ -37,13 +37,13 @@ interface Chat : LLM {
     promptStreaming(Prompt(question), context, conversationId, functions, promptConfiguration)
 
   @AiDsl
-  suspend fun promptStreaming(
```
raulraja, line 40: 👍

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
```diff
@@ -99,14 +108,17 @@ public CompletableFuture<List<String>> promptMessages(Chat llmModel, String prom
         return exec.future(continuation -> scope.promptMessages(llmModel, prompt, functions, promptConfiguration, continuation));
     }
 
+    public Publisher<String> promptStreaming(Chat gpt4all, String line, PromptConfiguration promptConfiguration) {
+        return ReactiveFlowKt.asPublisher(scope.promptStreaming(gpt4all, line, exec.getContext(), null, Collections.emptyList(), promptConfiguration));
```
raulraja, line 112:
```suggestion
        return ReactiveFlowKt.asPublisher(scope.promptStreaming(gpt4all, line, exec.getContext(), scope.conversationId, Collections.emptyList(), promptConfiguration));
```
This should be the current scope conversation id so the memory and previous messages are included in the prompt

raulraja (dismissed):
Pending fix for `conversationId`, looks great @Zevleg !

PR #257, opened 2023-07-19
implementation of CoreAIScope as Autocloseable

victorcrrd:
Little PR that adds an implementation of `CoreAIScope` as `AutoCloseable`. This implementation is based on @nomisRev implementation of the `AutoClose` DSL (see [here](https://gist.github.com/nomisRev/952c619fa40c7c0041974992a6ad1e06)). This way, we can add dependencies and resources with the `autoClose` function so that they will be automatically closed once the scope is closed.

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/auto/CoreAIScope.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/auto/CoreAIScope.kt
@@ -29,7 +29,7 @@ constructor(
   val embeddings: Embeddings,
   val context: VectorStore = LocalVectorStore(embeddings),
   val conversationId: ConversationId = ConversationId(UUID.generateUUID().toString())
-) : AutoCloseable, AutoClose by AutoClose() {
+) : AutoCloseable, AutoClose by AutoClose {
 
   /**
    * Allows invoking [AI] values in the context of this [CoreAIScope].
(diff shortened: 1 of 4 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/auto/CoreAIScope.kt
```diff
@@ -159,4 +161,22 @@ constructor(
     size: String = "1024x1024",
     promptConfiguration: PromptConfiguration = PromptConfiguration.DEFAULTS
   ): ImagesGenerationResponse = images(prompt, context, numberImages, size, promptConfiguration)
+
+  private val finalizers: Atomic<List<() -> Unit>> = Atomic(emptyList())
```
victorcrrd, line 165: I added the implementation here, but I don't know if it would make sense to have this implementation somewhere else. In that case, where should we have that? I guess that it should be in the `xef-core` module, but which package?

nomisRev: I would add it in a seperate file/class and use Kotlin delegation. So I would just copy the gist I shared 1-on-1 and then you can enable the syntax in this class using delegation.

```kotlin
: AutoCloseable, AutoClose by AutoClose() {
```

Well, almost 1-on-1. For the above syntax to work you'd need `AutoClose` to be an `interface` instead of a class and then have a `fun AutoClose(): AutoClose` smart-constructor which provides the implementation for the interface. I updated the gist, https://gist.github.com/nomisRev/952c619fa40c7c0041974992a6ad1e06.

nomisRev (approved):

PR #258, opened 2023-07-20
[Proposal] PromptConfiguration for different models

Montagon:
This PR implements a first idea of how we can have different `PromptConfiguration` for each type of model. In this way, we bring more versatility to users as they can configure each model with all the parameters.

### DIFF
--- a/examples/kotlin/src/main/kotlin/com/xebia/functional/xef/auto/gpt4all/LocalModels.kt
+++ b/examples/kotlin/src/main/kotlin/com/xebia/functional/xef/auto/gpt4all/LocalModels.kt
@@ -43,7 +43,9 @@ suspend fun main() {
 
     ai {
         gpt4All.use { gptModel: GPT4All ->
-            val promptConfiguration: PromptConfiguration = GPT4AllPromptConfiguration()
+            val promptConfiguration: PromptConfiguration = GPT4AllPromptConfiguration(
+                topK = 40
+            )
             val res =
                 gptModel.promptMessage(userInput, context = this.context, promptConfiguration = promptConfiguration)
             println("\n🤖 Done")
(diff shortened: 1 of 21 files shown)


gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/GPT4All.kt
```diff
@@ -29,148 +30,173 @@ import java.nio.file.Path
...
-                val c = b.toChar()
-                channel.trySend(c.toString())
-              }
+            val llModel = LLModel(path)
+
+
+            override suspend fun createCompletion(request: CompletionRequest): CompletionResult =
+                with(request) {
+                    val promptConfig: GPT4AllPromptConfiguration =
+                        request.promptConfiguration as GPT4AllPromptConfiguration
```
Montagon, line 87: I don't know if there is a better way on doing this

examples/kotlin/src/main/kotlin/com/xebia/functional/xef/auto/gpt4all/LocalModels.kt
```diff
@@ -0,0 +1,65 @@
...
+    println("🤖 GPT4All loaded: $gpt4All")
+
+//    val url = "https://gpt4all.io/models/ggml-gpt4all-j-v1.3-groovy.bin"
+
+    print("\n🤖 Enter the prompt: ")
+    val userInput = readlnOrNull() ?: return
+
+//    val prompt = "###User:\n$userInput\n###Assistant:"
+
+//    ai(model = gpt4All) { //this: GPT4AllAIScope
```
Montagon, line 36: This is just an idea of how this could be in the future

javipacheco:
It's important to find a solution to that. This is a good solution and I'm going to suggest another one, may be easier to implement at this point and we can find a better solution for the future when we have more local models working

Another approach could be to use a `Map` where the user can add all of the extra configurations in the model

```kotlin
class PromptConfiguration(
  val maxDeserializationAttempts: Int = 3,
  val user: String = Role.USER.name,
  val temperature: Double = 0.4,
  val numberOfPredictions: Int = 1,
  val docsInContext: Int = 5,
  val memoryLimit: Int = 5,
  val minResponseTokens: Int = 500,
  val streamToStandardOut: Boolean = false,
  val extras: Map<String, Any> = emptyMap(),
)
```

The user should know the `extras`, but this implementation could be a quick solution when we will add a new model with other configurations

Just another suggestion!

raulraja:
This proposal looks good so far 👍

PR #260, opened 2023-07-20
Bugs in conversations

javipacheco:
This PR tries to resolve 2 bugs:

## Bug 1

When we are working in conversations in Xef (remember that we are creating by default the `conversionId` in  `CoreAIScope`)

The problem is the order of the messages when we are sending the `messages` to `OpenAI`

Currently, we are storing the messages in the conversation in the wrong order. When we send the request to `OpenAI` the messages have the following order:

```
ASSISTANT: <response>
USER: <question>
ASSISTANT: <response>
USER: <question>
```

And when we add a new question from the user, we add this message to the end like this:

```
ASSISTANT: <response>
USER: <question>
ASSISTANT: <response>
USER: <question>
USER: <new-question>
```

In my tests, this is a problem that sometimes `OpenAI` isn't responding to the last question, it is responding to the second to last

The right order should be:

```
USER: <question>
ASSISTANT: <response>
USER: <question>
ASSISTANT: <response>
USER: <new-question>
```

This PR changes the way that we are storing the conversation in the memory in order to fix that

_Note: You can revert the changes in `Chat.kt` and you will see that the `Conversation.kt` example gives the same result both times_

## Bug 2

Resolving [this issue](https://github.com/xebia-functional/xef/issues/262)


Montagon (approved):

victorcrrd (dismissed):
I think the problem comes from [here](https://github.com/xebia-functional/xef/blob/main/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt#L276), where `memories` is being reversed. Those memories include, in some way, the messages that form the chat, and they are included in a way that older (question, answer) pairs go first, and questions go before answers. That is, `Q1, A1, ... Qn, An`. When reversing memories, we have `An, Qn, ..., A1, Q1`, so answers go before questions.

I don't know if it's better to reverse the order in which question, answer pairs are added to the memory (what you are doing) or if we should take into account that Chat memories come in pairs and that we should reverse them keeping the relative order of pairs the same (and thus changing that `messages` function.  I guess that for this case, your solution is better.

raulraja (approved):
my bad 🙋 , thank you!

javipacheco:
I hadn't noticed the `reversed()`, it's better to do it like this, thank you @victorcrrd 

I have found other 2 bugs in the conversations when we are storing the messages in the `LocalVectorStore`. I'm fixing them now

javipacheco:
I have updated the PR with the new fixes and added some tests

@victorcrrd @raulraja could you please review it?

core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
```diff
@@ -31,15 +31,27 @@ private constructor(private val embeddings: Embeddings, private val state: Atomi
...
-        orderedMemories = prevState.orderedMemories + memories.groupBy { it.conversationId }
+        orderedMemories = prevMemoriesWithNewMessages + allNewMemoriesWithoutPrevConversationId
       )
     }
   }
 
   override suspend fun memories(conversationId: ConversationId, limit: Int): List<Memory> {
     val memories = state.get().orderedMemories[conversationId]
-    return memories?.take(limit).orEmpty()
+    return memories?.takeLast(limit).orEmpty()
```
victorcrrd, line 54: @javipacheco and I have been talking, and I agree with this change to provide `memories` in the order they were created (that is, from oldest to newest), and getting (up to) the last `limit` ones. Now I'm wondering if we should consider this to happen for all possible implementations of `VectorStore`, providing users with the postcondition that `memories` always returns values ordered from oldest to newest and we return up to `limit` of them. Does that make sense, @raulraja? Is the intended usage/implementation different? In that case, we should revisit the other `VectorStore` implementations to make sure that that postconditions are always met. I propose it, and working on that on a different PR

javipacheco: I agree @victorcrrd, if this is the way to do that, we should review all of the implementations of `VectorStore` and we must create tests for every implementation

core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
```diff
@@ -31,15 +31,27 @@ private constructor(private val embeddings: Embeddings, private val state: Atomi
 
   override suspend fun addMemories(memories: List<Memory>) {
```
victorcrrd, line 32: A shorter option, but I don't know if less readable:
```kotlin
override suspend fun addMemories(memories: List<Memory>) {
    state.update { prevState ->
      prevState.copy(
        orderedMemories = memories.groupBy { it.conversationId }.let { memories ->
          (prevState.orderedMemories.keys + memories.keys).associateWith { key ->
            val l1 = prevState.orderedMemories[key] ?: emptyList()
            val l2 = memories[key] ?: emptyList()
            l1 + l2
          }
        }
      )
    }
  }
```

core/src/commonTest/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStoreSpec.kt
```diff
@@ -0,0 +1,34 @@
...
+            messages shouldBe messagesExpected
+        }
+
+        "memories function should return the last n messages in the right order" {
+            val localVectorStore = LocalVectorStore(FakeEmbeddings())
+
+            localVectorStore.addMemories(messages1)
+            localVectorStore.addMemories(messages2)
+
+            val messages = localVectorStore.memories(defaultConversationId, 2)
```
victorcrrd, line 27: I think I would define a value for that `2` you are using as the `limit` to be used also in the `takeLast` function

core/src/commonTest/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStoreSpec.kt
victorcrrd, line 1: I would also add a test with messages from a different conversation to test how we are combining them

examples/kotlin/src/main/kotlin/com/xebia/functional/xef/auto/Conversation.kt
```diff
@@ -0,0 +1,48 @@
...
+    ai {
+
+        val email: String =
+            promptMessage(
+                """
+                |You are a Marketing Responsible and have the information about different products. You have to prepare 
+                |an email template with the personal information
+            """.trimMargin()
+            )
+        println("First question:\n $email")
```
victorcrrd, line 16: I would print first the prompt and then the answer in all three calls to the AI to show it like if it was a chat conversation

core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
javipacheco: My opinion is: This is great!

core/src/commonTest/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStoreSpec.kt
javipacheco: Yes, I'm creating new tests for that

core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
raulraja: We just need to make sure we are converting the memories into messages in the appropriate order as the chat completion endpoint expects them. See open ai api docs for details of what the conversation looks like in terms of order

javipacheco: The [OpenAI documentation](https://platform.openai.com/docs/api-reference/chat/create) does not explicitly state that the order of messages has to be in ascending order, although [all examples](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb) show that they should be sent in ascending order.

I have been doing some tests in postman and when I change the order of messages, I receive different results. 

If we want to change all of the implementations of `VectorStore` we need to change the following:

- `CombinedVectorStore`
- `Lucene`
- `PGVectorStore`: maybe we need `test-containers` for creating tests in this store

I think that we should address this in separated PRs

raulraja: Let's do this in this PR. Let me know if there is anything I can do to help

javipacheco:
@victorcrrd @raulraja I have fixed some problems in the order messages on `CombinedVectorStore` and I have created some tests for this

We are using `CombinedVectorStore` in `contextScope` inside Xef. They are important changes in the base code. Could you please review the changes?

core/src/commonTest/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStoreSpec.kt
victorcrrd, line 1: All I can think of is Property-Based Testing 👀

core/src/commonTest/kotlin/com/xebia/functional/xef/vectorstores/CombinedVectorStoreSpec.kt
victorcrrd, line 1: I think we should also test adding memories to a combined vector store. Here we are creating a `CombinedVectorStore` directly from two `VectorStore`s, but we are not adding messages directly through the `CombinedVectorStore` API

raulraja (approved):

javipacheco:
@raulraja @Montagon This PR is ready for review. Here are the changes: 

- When we are getting messages from the `VectorStore` implementations, the messages are sorted by timestamp
- Messages are always sent in ascending order
- We should address the `Lucene` implementation in a separate PR. We have to include integrations tests for that and this is not the proposal for this PR

raulraja (approved):
thank you!

PR #261, opened 2023-07-20
#861n30d7t: Making OpenAI AutoCloseable and AutoClose

victorcrrd:
Work towards better handling the lifetime of classes and the resources they need. In this case, this PR implements the `OpenAI` class (that provides a number of LLM to be easily consumed by users) as `AutoCloseable` and `AutoClose`, making sure that all `OpenAIClient`s are closed when we are no longer using the model.

### DIFF
--- a/openai/src/commonMain/kotlin/com/xebia/functional/xef/auto/llm/openai/OpenAI.kt
+++ b/openai/src/commonMain/kotlin/com/xebia/functional/xef/auto/llm/openai/OpenAI.kt
@@ -3,39 +3,43 @@ package com.xebia.functional.xef.auto.llm.openai
 import arrow.core.nonEmptyListOf
 import com.xebia.functional.tokenizer.ModelType
 import com.xebia.functional.xef.AIError
+import com.xebia.functional.xef.auto.AutoClose
+import com.xebia.functional.xef.auto.autoClose
 import com.xebia.functional.xef.env.getenv
 import kotlin.jvm.JvmField
 
-class OpenAI(internal val token: String) {
-  val GPT_4 = OpenAIModel(this, "gpt-4", ModelType.GPT_4)
+class OpenAI(internal val token: String) : AutoCloseable, AutoClose by autoClose() {
+  val GPT_4 by lazy { autoClose(OpenAIModel(this, "gpt-4", ModelType.GPT_4)) }
 
-  val GPT_4_0314 = OpenAIModel(this, "gpt-4-0314", ModelType.GPT_4)
+  val GPT_4_0314 by lazy { autoClose(OpenAIModel(this, "gpt-4-0314", ModelType.GPT_4)) }
 
-  val GPT_4_32K = OpenAIModel(this, "gpt-4-32k", ModelType.GPT_4_32K)
+  val GPT_4_32K by lazy { autoClose(OpenAIModel(this, "gpt-4-32k", ModelType.GPT_4_32K)) }
 
-  val GPT_3_5_TURBO = OpenAIModel(this, "gpt-3.5-turbo", ModelType.GPT_3_5_TURBO)
+  val GPT_3_5_TURBO by lazy { autoClose(OpenAIModel(this, "gpt-3.5-turbo", ModelType.GPT_3_5_TURBO)) }
 
-  val GPT_3_5_TURBO_16K = OpenAIModel(this, "gpt-3.5-turbo-16k", ModelType.GPT_3_5_TURBO_16_K)
+  val GPT_3_5_TURBO_16K by lazy { autoClose(OpenAIModel(this, "gpt-3.5-turbo-16k", ModelType.GPT_3_5_TURBO_16_K)) }
 
-  val GPT_3_5_TURBO_FUNCTIONS =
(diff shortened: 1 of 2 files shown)


openai/src/commonMain/kotlin/com/xebia/functional/xef/auto/llm/openai/OpenAI.kt
```diff
@@ -44,6 +48,8 @@ class OpenAI(internal val token: String) {
         ?: throw AIError.Env.OpenAI(nonEmptyListOf("missing OPENAI_TOKEN env var"))
     }
 
+    // TODO: Handle the lifetime of these values!
+
     @JvmField val DEFAULT = OpenAI(openAITokenFromEnv())
```
victorcrrd, line 53: Right now I'm trying to figure out how the lifetime of a companion object is handled. All `OpenAIModel` values defined below are tied to the lifetime of this `DEFAULT` instance of `OpenAI`, so they will correctly be closed when `DEFAULT` is closed (and, thus, closing the OpenAI `client`). However, `DEFAULT` is instantiated in the companion object of the class and I have to make sure that it's closed somewhere (or make sure that the `OpenAIModel` instances it defines are not initialised unless actually being used

victorcrrd: The existence of an instance to be used by default implies that it's lifetime is the whole's application's lifetime. That is, we either have a default instance built at the beginning of the application, or we can make sure its lifetime is as short as needed. So I guess that this PR should end here, with future work dealing with how we can provide users with default AI models to use.

Montagon (dismissed):

nomisRev (approved):

openai/src/commonMain/kotlin/com/xebia/functional/xef/auto/llm/openai/OpenAI.kt
```diff
@@ -3,39 +3,43 @@ package com.xebia.functional.xef.auto.llm.openai
...
+  val TEXT_ADA_001 by lazy { autoClose(OpenAIModel(this, "text-ada-001", ModelType.TEXT_ADA_001)) }
 
-  val TEXT_EMBEDDING_ADA_002 =
-    OpenAIModel(this, "text-embedding-ada-002", ModelType.TEXT_EMBEDDING_ADA_002)
+  val TEXT_EMBEDDING_ADA_002 by lazy {
+    autoClose(OpenAIModel(this, "text-embedding-ada-002", ModelType.TEXT_EMBEDDING_ADA_002))
+  }
 
-  val DALLE_2 = OpenAIModel(this, "dalle-2", ModelType.GPT_3_5_TURBO)
+  val DALLE_2 by lazy { autoClose(OpenAIModel(this, "dalle-2", ModelType.GPT_3_5_TURBO)) }
```
raulraja, line 42: The issue with autoclosing resources declared in the companion like this, whose lifetime is the entire application and not initialized with a function, is that once you close them, they will become unusable to be reopened since the companion state will point to a stale object. You'd need to change all these values to function or perform the Autoclosing on JVM shutdown hooks that await termination.

Maybe I misunderstood `autoClose,` and this is safe, but when you run this through a scope, once the Scope ends, this object will become stale for the remainder of the application.

In practice, the resources this object holds are just the underlying OpenAI or similar HTTP clients, which don't do persistence or anything relevant on shutdown. \do they need to be closed and created on each Scope?, or can we have these objects for the entire Application lifecycle ? //cc @nomisRev

nomisRev: > The issue with autoclosing resources declared in the companion like this

These are defined in the `OpenAI` class not the `companion object`? 🤔 I think you understood `autoClose` correct, it's basically the same as `autoCloseable { }`, or `install`, inside `ResourceScope`.

Your rationale applies to the `DEFAULT`, `DEFAULT_CHAT`, etc instances defined below in the `companion object`  which creates instances of the `OpenAI` class holding these other instances. The `DEAULT_X` instances are currently still leaking. Which are currently not _closing_ only any scope and thus leaking. It's currently not scoped to the application scope, or the `AIScope` in any way. Which I think would be ideal, but some changes in the entire `xef` library seem to be needed for that.

> which don't do persistence or anything relevant on shutdown. \do they need to be closed and created on each Scope?

I would say yes, because they are defined as `Closeable` themselves. Typically they will hold some `ExecutorService`, or `CoroutineScope`, and often a bunch of other stateful things inside. Especially when composing different blocks together, which might have different scopes I'd say it become more important. In small examples that, or single scope applications, you won't face any issues.

Another issue is that currently we're hardcoding to `openAITokenFromEnv` given that these instances are singletons, and thus we cannot provide custom configurations. This is a roadblock for web for example.

raulraja: Makes sense @nomisRev resolving this and we can address those issues in separate PRs

raulraja (approved):

PR #271, opened 2023-07-25
Fix Ktor engine

nomisRev:
- Adds missing client for JS in OpenAI module
- moves JVM engine dependency from core to OpenAI module.
  - `jvmMain` was relying on `io.ktor.http` for `BingSearch` scrapping, with was coming indirectly from the JVM CIO engine.
  - `xef-pdf` was relying on the transitive dependency of JVM CIO engine through `xef-core`.

### DIFF
--- a/openai/build.gradle.kts
+++ b/openai/build.gradle.kts
@@ -80,14 +80,19 @@ kotlin {
     val jvmMain by getting {
       dependencies {
         implementation(libs.logback)
+        api(libs.ktor.client.cio)
       }
     }
-
-    val jsMain by getting
 
     val jvmTest by getting {
       dependencies {
         implementation(libs.kotest.junit5)
+      }
+    }
+
+    val jsMain by getting {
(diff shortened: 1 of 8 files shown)


openai/build.gradle.kts
```diff
@@ -80,17 +80,22 @@ kotlin {
     val jvmMain by getting {
       dependencies {
         implementation(libs.logback)
+        api(libs.ktor.client.cio)
```
raulraja, line 83: Does the OpenAI module need any ktor deps, or does it get it directly through the client library dependency?

nomisRev: It's written in an abstract way, the engines are packaged separately so need to be explicitly imported by a downstream project.

raulraja: thanks @nomisRev !

raulraja (dismissed):

raulraja (approved):

PR #272, opened 2023-07-25
PGVectorStore tests

victorcrrd:
Old tests were not passing correctly, so I solved that. Also solved the bug in `PGVectorStore` and added a very simple test. More details about important changes in the comments

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/Memory.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/Memory.kt
@@ -2,4 +2,11 @@ package com.xebia.functional.xef.vectorstores
 
 import com.xebia.functional.xef.llm.models.chat.Message
 
+/**
+ * Representation of the memory of a message in a conversation.
+ *
+ * @property content message sent.
+ * @property conversationId uniquely identifies the conversation in which the message took place.
+ * @property timestamp in milliseconds.
+ */
 data class Memory(val conversationId: ConversationId, val content: Message, val timestamp: Long)
--- a/integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/PostgreSQLVectorStore.kt
+++ b/integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/PostgreSQLVectorStore.kt
@@ -83,9 +83,9 @@ class PGVectorStore(
 
   fun createCollection(): Unit =
     dataSource.connection {
-      val xa = UUID.generateUUID()
+      val uuid = UUID.generateUUID()
       update(addNewCollection) {
-        bind(xa.toString())
+        bind(uuid.toString())
         bind(collectionName)
       }
     }
--- a/integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/PostgreSQLVectorStore.kt
+++ b/integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/PostgreSQLVectorStore.kt
@@ -51,7 +51,7 @@ class PGVectorStore(
           content = Message(
             role = Role.valueOf(role.uppercase()),
             content = content,
-            name = "role",
+            name = role,
           ),
           timestamp = timestamp,
         )
--- a/integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/postgresql/postgres.kt
+++ b/integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/postgresql/postgres.kt
@@ -23,7 +23,7 @@ val createMemoryTable: String =
        conversation_id TEXT NOT NULL,
        role TEXT NOT NULL,
        content TEXT UNIQUE NOT NULL,
-       timestamp TIMESTAMP NOT NULL,
+       timestamp TIMESTAMP NOT NULL
      );"""
     .trimIndent()
 
@@ -97,7 +97,7 @@ val getCollectionById: String =
 val getMemoriesByConversationId: String =
   """SELECT * FROM xef_memory
      WHERE conversation_id = ?
-     ORDER BY timestamp DESC LIMIT ?;"""
+     ORDER BY timestamp DESC LIMIT ?;""" // DESC -> ASC
     .trimIndent()
 
 val addNewDocument: String =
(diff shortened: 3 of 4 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/Memory.kt
```diff
@@ -2,4 +2,11 @@ package com.xebia.functional.xef.vectorstores
 
 import com.xebia.functional.xef.llm.models.chat.Message
 
+/**
+ * Representation of the memory of a message in a conversation.
+ *
+ * @property content message sent.
+ * @property conversationId uniquely identifies the conversation in which the message took place.
+ * @property timestamp in milliseconds.
```
victorcrrd, line 10: This is the important part, but I guess it's not that important now. The issue was that we are modelling `Memories` to have a `timestamp` of type `Long`. That value may be stored in a postgres database, and the schema required that field to be of type `TIMESTAMP`, so I needed to make sure how to transform from a `Long` to a `TIMESTAMP`. Right now that is avoided, because I changed the schema to use `BIGINT` instead of `TIMESTAMP`. More about that below

integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/PostgreSQLVectorStore.kt
```diff
@@ -51,7 +51,7 @@ class PGVectorStore(
           content = Message(
             role = Role.valueOf(role.uppercase()),
             content = content,
-            name = "role",
+            name = role,
```
victorcrrd, line 54: We were using always `"role"` as the name in every `Message` read

integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/postgresql/postgres.kt
```diff
@@ -23,7 +23,7 @@ val createMemoryTable: String =
        conversation_id TEXT NOT NULL,
        role TEXT NOT NULL,
        content TEXT UNIQUE NOT NULL,
-       timestamp TIMESTAMP NOT NULL,
+       timestamp BIGINT NOT NULL
```
victorcrrd, line 26: Here I'm changing the `xef_memories` schema from using `TIMESTAMP` to `BIGINT`. I thought that it wasn't very important for this value to be a `TIMESTAMP`, we just needed something that could be ordered. Maybe for performance issues it's better to have this as a `TIMESTAMP` instead of a `BIGINT`. However, having it as a `TIMESTAMP` here and modelling it in the application as a `Long` required translating back and forth, which was more difficult. Let me know your opinion on that and if you agree I can change it and implement both translations

integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/postgresql/postgres.kt
```diff
@@ -97,7 +97,7 @@ val getCollectionById: String =
 val getMemoriesByConversationId: String =
   """SELECT * FROM xef_memory
      WHERE conversation_id = ?
-     ORDER BY timestamp DESC LIMIT ?;"""
+     ORDER BY timestamp ASC LIMIT ?;"""
```
victorcrrd, line 100: This was the actual change we needed!

integrations/postgresql/src/test/kotlin/xef/PGVectorStoreSpec.kt
```diff
@@ -23,12 +29,12 @@ val postgres: PostgreSQLContainer<Nothing> =
 
 class PGVectorStoreSpec :
   StringSpec({
-    val container = install(SharedTestContainerExtension(postgres))
```
victorcrrd, line 26: `SharedTestContainerExtension` was deprecated in favor of `ContainerExtension`

integrations/postgresql/src/test/kotlin/xef/PGVectorStoreSpec.kt
```diff
@@ -23,12 +29,12 @@ val postgres: PostgreSQLContainer<Nothing> =
...
+    val container = install(ContainerExtension(postgres))
     val dataSource =
       autoClose(
         HikariDataSource(
           HikariConfig().apply {
-            jdbcUrl = container.jdbcUrl
+            jdbcUrl = container.jdbcUrl.replace("localhost", "0.0.0.0") // workaround: needs to be improved
```
victorcrrd, line 37: I had trouble executing the tests: they couldn't connect to the database because using `localhost` as the host causes that the container cannot be found. This should be solved somehow else, let me know any ideas!

integrations/postgresql/src/test/kotlin/xef/PGVectorStoreSpec.kt
```diff
@@ -63,6 +69,10 @@ class PGVectorStoreSpec :
 
     "createCollection should create collection" { pg.createCollection() }
 
+    "addTexts should not fail now that we created the collection" {
+      pg.addTexts(listOf("foo", "bar"))
```
victorcrrd, line 73: texts were not added, so all tests after `"createCollection should create collection"` that used embeddings were failing. This was needed!

integrations/postgresql/src/test/kotlin/xef/PGVectorStoreSpec.kt
```diff
@@ -78,6 +88,19 @@ class PGVectorStoreSpec :
     "similaritySearchByVector should return document" {
       pg.similaritySearchByVector(Embedding(listOf(1.0f, 2.0f, 3.0f)), 1) shouldBe listOf("foo")
     }
+
+    "memories added in order should be obtained in the same order" {
```
victorcrrd, line 92: Test I added for messages

integrations/postgresql/src/test/kotlin/xef/PGVectorStoreSpec.kt
raulraja: Maybe just use 0.0.0.0 in the test. We should not handle address resolution problems ourselves.

raulraja (approved):
👏 @victorcrrd

integrations/postgresql/src/test/kotlin/xef/PGVectorStoreSpec.kt
victorcrrd: Then I leave it like that!

javipacheco (approved):
Great work!! LGTM!

PR #277, opened 2023-07-26
Split jdk8 jdk21

Zevleg:
Ticket [Split Java examples into 2 modules. JDK8 and JDK 21](https://app.clickup.com/t/861n405en)


raulraja (approved):
Looks great @Zevleg .

I think going forward, we have to think of a strategy that isn't repeating all examples.
For the jdk21 and others with LOOM support we should have `blocking` version of the methods in addition to those that return CompletableFuture. 

We are investigating in a different spike how to get rid of most of the Java AIScope wrapping and have those methods exported directly from the xef-core JVM version. If we are able to do that then we can stop wrapping the Kotlin methods in the Scala and Java impls.

anamariamv (approved):

PR #283, opened 2023-07-28
Updated task to run spotless

Montagon:
I think that the `project.exec` block only executes the last `commandLine` that was set instead of running the two commands. This PR updates it to have two different `exec` blocks.


raulraja:
@Montagon looks like the win check is failing with:
```
* What went wrong:
Could not determine the dependencies of task ':xef-scala-cats:spotlessCheck'.
> Could not create task ':xef-scala-cats:spotlessScalaCheck'.
   > Could not create task ':xef-scala-cats:spotlessScala'.
      > Could not resolve all dependencies for configuration ':xef-scala-cats:spotless1625108215'.
         > Failed to calculate the value of task ':xef-scala-cats:compileJava' property 'javaCompiler'.
            > No matching toolchains found for requested specification: {languageVersion=11, vendor=any, implementation=vendor-specific}.
               > No locally installed toolchains match (see https://docs.gradle.org/8.1/userguide/toolchains.html#sec:auto_detection) and toolchain download repositories have not been configured (see https://docs.gradle.org/8.1/userguide/toolchains.html#sub:download_repositories).
```

raulraja:
@Montagon, is this still relevant, or can we close it?

Montagon:
> @Montagon, is this still relevant, or can we close it?

I think we can close it for now. Not sure if it's fixed with the new lint options.

PR #284, opened 2023-07-28
Add GCP Chat with example

nomisRev:


raulraja (dismissed):
It looks great @nomisRev !.

We need to also support the Embeddings wired to the GCP embeddings models. Currently, the `getOrThrow` imported in the example brings the embeddings from the GPT4All module which are actually not working great and depend on an adapter from dl java for HuggingFace.

I think the way we are using single embeddings generically for all documents in the `ai` while we have different models that may require tokenizing differently may not yield always the best results.

nomisRev:
@raulraja do you want me to change anything in this PR, or shall we pick it up in subsequent PRs?

integrations/gcp/src/commonMain/kotlin/com/xebia/functional/xef/gcp/GcpChat.kt
```diff
@@ -0,0 +1,109 @@
...
+  private fun List<Message>.buildPrompt(): String {
+    val messages: String =
+      joinToString("") { message ->
+        when (message.role) {
+          Role.SYSTEM -> message.content
+          Role.USER -> "\n### Human: ${message.content}"
+          Role.ASSISTANT -> "\n### Response: ${message.content}"
+        }
+      }
+    return "$messages\n### Response:"
```
Montagon, line 107: I'm thinking if this is mandatory or not 🤔 

From [here](https://cloud.google.com/vertex-ai/docs/generative-ai/chat/chat-prompts) we can see that the format should be something like this:
```json
"messages": [
  {
    "author": "USER",
    "content": "Hi",
  },
  { 
    "author": "AI",
    "content": "Hi! How can I help you today?",
  },
  { 
    "author": "USER",
    "content": "I just want to talk.",
  },
  { 
    "author": "AI",
    "content": "Sure, I'm here to listen. What would you like to talk about?",
  },
  { 
    "author": "USER",
    "content": "Let's talk about movies.",
  }],
  ```

File: integrations/gcp/src/commonMain/kotlin/com/xebia/functional/xef/gcp/GcpChat.kt [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+import kotlinx.uuid.generateUUID
+
+@OptIn(ExperimentalStdlibApi::class)
+class GcpChat(apiEndpoint: String, projectId: String, modelId: String, token: String) :
+  Chat, Completion, AutoCloseable {
+  private val client: GcpClient = GcpClient(apiEndpoint, projectId, modelId, token)
+
+  override val name: String = client.modelId
+  override val modelType: ModelType =
+    ModelType.LocalModel(client.modelId, EncodingType.CL100K_BASE, 4096)
```
Montagon, line 25: The model used in the example only has 2048 as max tokens. I don't know if is better to set here the value. I don't know if is possible to modify from another place: 
```suggestion
    ModelType.LocalModel(client.modelId, EncodingType.CL100K_BASE, 2048)
```

Montagon (approved):
Thanks Simon! Left only two comments

integrations/gcp/src/commonMain/kotlin/com/xebia/functional/xef/gcp/GcpChat.kt
nomisRev: Let’s revisit this in a later PR. Is that okay with you?

Montagon: Yes, sure! 👍

Montagon (approved):

Issue #275, opened 2023-07-26
RequestConfig not constructable from other jvm languages


jackcviers:
Kotlin data-classes cannot be constructed in Java or Scala. To fix this, the data class constructor must be annotated with `@JvmOverloads`.

Without this it is not possible to create a `PGVectorStore`: https://github.com/xebia-functional/xef/blob/e1561909b1c94fba3d6b9eabc0f5d44e433ccdb4/integrations/postgresql/src/main/kotlin/com/xebia/functional/xef/vectorstores/PostgreSQLVectorStore.kt#L20

https://github.com/xebia-
functional/xef/blob/e1561909b1c94fba3d6b9eabc0f5d44e433ccdb4/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/embeddings/RequestConfig.kt#L5-L8

sbrunk:
I just ran into this as well while trying out xef from Scala.

`RequestConfig` wants a (probably synthetically generated?) `kotlin.jvm.internal.DefaultConstructorMarker` as third parameter, which can't be instantiated. Adding `@JvmOverloads` to the constructor did not work for me.

I also tried adding secondary `constructor(model: EmbeddingModel, user: String): this(model, User(user))` but then the Kotlin compiler complains that both constructors have the same JVM signature.

As a workaround, I added a static factory like this. I'm sure there's a better way, but my Kotlin knowledge is still quite limited.

```kotlin
data class RequestConfig(val model: EmbeddingModel, val user: User) {
  companion object {
    @JvmStatic
    fun create(model: EmbeddingModel, userId: String): RequestConfig {
      return RequestConfig(model, User(userId))
    }
    @JvmInline value class User(val id: String)
  }
```

jackcviers:
Make a PR @sbrunk with a test in xef-scala, please!

sbrunk:
@jackcviers done in #285

raulraja:
These changes should be available in [0.0.3-alpha.2](https://github.com/xebia-functional/xef/tree/0.0.3-alpha.2)

PR #285, opened 2023-07-28
Add static apply method to enable creating RequestConfig from Scala

sbrunk:
With the factory, it should be possible to create `RequestConfig` instances from Scala and Java.

This seems necessary because `User` is an inline value class. See #275

I'm not sure if it also makes sense to add a String getter for user.


raulraja (approved):
Thank you @sbrunk !
I took the liberty to merge main back into your branch and apply spotless so the build would pass. 

I'm guessing there are more places like this one since we have not yet done a pass to make all the classes and functions compatible for all JVM langs but Kotlin. The same applies to `suspend` functions which we also plan to address in a better way than what we are doing now wrapping them.

raulraja:
this change will be available since [0.0.3-alpha.2](https://github.com/xebia-functional/xef/tree/0.0.3-alpha.2)

sbrunk:
Thanks @raulraja! I'll continue playing with it :)

PR #289, opened 2023-08-01
Add SerpApi Example

Zevleg:
Ticket [Implement Search API](https://app.clickup.com/t/85ztn9hk8)

Output result:

image


reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
```diff
@@ -0,0 +1,71 @@
...
+                }
+            )
+        }
+    }
+
+    @Serializable
+    data class SearchData(
+        val q: String,
+        val location: String? = null,
+        val hl: String? = null,
```
raulraja, line 38: We can use here `@SerialName("hl") betterNameHere: String? = null` to have properties serialized as expected but still be able to make sense of them when we read them in kotlin.

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
```diff
@@ -0,0 +1,71 @@
...
+        val gl: String? = null,
+        val googleDomain: String = "google.com",
+        val apiKey: String
+    )
+
+    @Serializable data class SearchResults(@JsonNames("organic_results") val searchResult: List<SearchResult>)
```
raulraja, line 44: For better compatibility with Java, every class should be on its own file.

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
```diff
@@ -0,0 +1,71 @@
...
+                Json {
+                    encodeDefaults = false
+                    isLenient = true
+                    ignoreUnknownKeys = true
+                }
+            )
+        }
+    }
+
+    @Serializable
```
raulraja, line 34: Does this ever get serialized? It looks like the `SearchData` properties are used to construct a query string manually from its properties, which means this class does not need to be `@Serializable`

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
```diff
@@ -0,0 +1,71 @@
...
+        val response =
+            http.get(
+                "https://serpapi.com/search.json?q=${searchData.q}&location=${searchData.location}&hl=${searchData.hl}&gl=${searchData.gl}&google_domain=${searchData.googleDomain}&api_key=${searchData.apiKey}"
+            ) {
+                contentType(ContentType.Application.Json)
+            }
+
+        return if (response.status.isSuccess())
+            response.body<SearchResults>()
+        else throw SerpApiClientException(response.status, response.bodyAsText())
```
raulraja, line 62: We probably don't want to throw an exception here for some HTTP error codes. In some cases here we want to retry with some kind of exponential backoff. https://ktor.io/docs/client-retry.html#conditions

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
```diff
@@ -0,0 +1,71 @@
...
+        val title: String,
+        @JsonNames("snippet") val document: String,
+        val link: String
+    )
+
+    suspend fun search(searchData: SearchData): SearchResults {
+
+        val response =
+            http.get(
+                "https://serpapi.com/search.json?q=${searchData.q}&location=${searchData.location}&hl=${searchData.hl}&gl=${searchData.gl}&google_domain=${searchData.googleDomain}&api_key=${searchData.apiKey}"
```
raulraja, line 55: The `apiKey` should probably be in the `SerpApiClient` constructor as a nullable `String?`, if `null` is found, then we should load it from the environment with `getenv("SERP_API_KEY")` or similar. We don't want callers of this service to have to pass the key each time they create a request.

raulraja (changes requested):
Looks great, but I think we should address the comments before merging so this class is consistent with how the other classes and services work with credentials, error handling, and names.

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
Zevleg: Done

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
Zevleg: Done

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
Zevleg: Done

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
Zevleg: Done

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
```diff
@@ -0,0 +1,80 @@
...
+import io.ktor.client.plugins.*
+import io.ktor.client.plugins.contentnegotiation.*
+import io.ktor.client.request.*
+import io.ktor.http.*
+import io.ktor.serialization.kotlinx.json.*
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+
+class SerpApiClient : AutoCloseable {
```
raulraja, line 15: should implement `AutoClose by autoClose()` if we want it to close automatically with the `ai` block

raulraja (dismissed):
looks great, added a final comment to ensure it autocloses

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/serpapi/SerpApiClient.kt
Zevleg: it has override close of autoClose() like in GcpClient.kt

```kotlin
override fun close() {
        http.close()
    }
``` 
it seems that implement is for abstract

image

Zevleg: Done

raulraja (approved):
Looks great, thank you @Zevleg !

PR #298, opened 2023-08-04
Expression Language for LLM driven template replacements

raulraja:
The `Expression` class is a utility class designed to simplify the process of creating and running expressions for a chatbot or conversational AI system. It provides a fluent API for building expressions with different roles (SYSTEM, USER, and ASSISTANT) and generating dynamic prompts with variable keys.

Here's a breakdown of the key components and functionalities of the `Expression` class:

1. Constructor:
   - The `Expression` class takes three parameters: `scope`, `model`, and a `block` function. `scope` is an instance of `CoreAIScope`, `model` is an instance of `ChatWithFunctions`, and `block` is a lambda function that takes a `suspend Expression.() -> Unit` as its parameter. The `block` function is used to define the sequence of messages and prompts within the expression.

2. Logging:
   - The class uses the `KotlinLogging` library to create an instance of `KLogger` for logging purposes.

3. Messages and Generation Keys:
   - The class has two mutable lists: `messages` and `generationKeys`. The `messages` list stores messages with different roles (SYSTEM, USER, ASSISTANT) that are added using the `system`, `user`, and `assistant` functions. The `generationKeys` list keeps track of variable keys used in the prompts, which are added using the `prompt` function.

4. Message Builders:
   - The `system`, `user`, and `assistant` functions are used to add messages to the `messages` list with the corresponding roles.

5. Prompt Generation:
   - The `prompt` function is used to generate prompts with variable keys (e.g., {{$key}}) and store the variable keys in the `generationKeys` list.

6. `run` Function:
   - The `run` function executes the expression defined in the `block` function. It builds an instruction message using an `ExpertSystem`, adds it to the `messages` list, and then runs the prompt messages using the `model.prompt` function. The function also replaces the variable keys in the prompts with actual values obtained from the chatbot response.

7. `companion object run` Function:
   - The `run` function within the `companion object` allows for a concise way of creating and executing an `Expression` instance. It is a convenience function that takes a `scope`, `model`, and a `block` lambda as arguments, and it automatically calls the `run` function on the `Expression` instance.

8. New methods for the `Chat` interface that can consume a list of `Message` instead of a `String` or a `Prompt`

The code example below demonstrates how to use the `Expression` class to create a workout plan based on user input and the responses generated by the assistant.

```kotlin
package com.xebia.functional.xef.auto.expressions

import com.xebia.functional.xef.auto.CoreAIScope
import com.xebia.functional.xef.auto.ai
import com.xebia.functional.xef.auto.llm.openai.getOrThrow
import com.xebia.functional.xef.llm.ChatWithFunctions
import com.xebia.functional.xef.prompt.expressions.Expression
import com.xebia.functional.xef.prompt.expressions.ExpressionResult

suspend fun workoutPlan(
  scope: CoreAIScope,
  model: ChatWithFunctions,
  goal: String,
  experienceLevel: String,
  equipment: String,
  timeAvailable: Int
): ExpressionResult = Expression.run(scope = scope, model = model, block = {
  system { "You are a personal fitness trainer" }
  user {
    """
     |I want to achieve $goal.
     |My experience level is $experienceLevel, and I have access to the following equipment: $equipment.
     |I can dedicate $timeAvailable minutes per day.
     |Can you create a workout plan for me?
  """.trimMargin()
  }
  assistant {
    """
     |Sure! Based on your goal, experience level, equipment available, and time commitment, here's a customized workout plan:
     |${prompt("workout_plan")}
  """.trimMargin()
  }
})

suspend fun main() {
  val model = OpenAI.DEFAULT_SERIALIZATION
  ai {
    val plan = workoutPlan(
      scope = this,
      model = model,
      goal = "building muscle",
      experienceLevel = "intermediate",
      equipment = "dumbbells, bench, resistance bands",
      timeAvailable = 45
    )
    println("--------------------")
    println("Workout Plan")
    println("--------------------")
    println("🤖 replaced: ${plan.values.replacements.joinToString { it.key }}")
    println("--------------------")
    println(plan.result)
    println("--------------------")
  }.getOrThrow()
}
```
```text
--------------------
Workout Plan
--------------------
🤖 replaced: workout_plan
--------------------
You are a personal fitness trainer
I want to achieve building muscle.
My experience level is intermediate, and I have access to the following equipment: dumbbells, bench, resistance bands.
I can dedicate 45 minutes per day.
Can you create a workout plan for me?
Sure! Based on your goal, experience level, equipment available, and time commitment, here's a customized workout plan:
Here's a sample workout plan for you:

Day 1: Upper Body
- Dumbbell bench press: 3 sets of 10 reps
- Bent-over rows with dumbbells: 3 sets of 10 reps
- Dumbbell shoulder press: 3 sets of 10 reps
- Bicep curls with resistance bands: 3 sets of 10 reps

Day 2: Lower Body
- Dumbbell squats: 3 sets of 10 reps
- Dumbbell lunges: 3 sets of 10 reps
- Romanian deadlifts with dumbbells: 3 sets of 10 reps
- Calf raises with resistance bands: 3 sets of 10 reps

Day 3: Full Body
- Dumbbell chest press: 3 sets of 10 reps
- Dumbbell rows: 3 sets of 10 reps
- Dumbbell shoulder press: 3 sets of 10 reps
- Dumbbell squats: 3 sets of 10 reps

Remember to warm up before each workout and cool down/stretch afterwards. Increase the weight/resistance as you get stronger and aim to progressively overload your muscles. Rest for 1-2 minutes between sets. Stay consistent and have fun with your workouts!

--------------------
```


core/src/commonMain/kotlin/com/xebia/functional/xef/prompt/expressions/Expression.kt
```diff
@@ -0,0 +1,98 @@
...
+          |<template>
+          |${messages.joinToString("\n") { it.content }}
+          |</template>
+          |The variables are:
+          |${generationKeys.joinToString("\n") { it }}
+        """
+                  .trimMargin(),
+              instructions =
+                listOf(
+                  "Create a `ReplaceKeys` object `replacements` property of type `Map<String, String>` where the keys are the variable names and the values are the values to replace them with.",
```
jackcviers, line 63: Suggestion:

Instead of using Map String String and single value replacement, why not use an expert in a popular templating language, such as Handlebars? This would allow for more complex json object substitutions, loops for lists, etc.

As long as the templating language is popular enough, I think the LLM will be able to come up with a valid template for it.

raulraja: I tried the handlebars syntax support but the LLM is more precise when I ask it to place the structure back on the map than when I ask it to replace all the placeholders in the text, but with some prompt tweaking, it would support handlebars syntax. I still like that we use functions like `prompt` to place the placeholders where generation would happen instead of just providing an entire template and asking it to infer the keys. This allows us to save the keys and ask directly to translate those keys. We could do that with handlebars and some parsing too. Still, I also think you can support the handlebars' control structures and syntax simultaneously if we provide a way to know the keys and regions before we ask the LLM. If you'd like to play with it in the same branch we can try to evolve it to have more powerful templating.

jackcviers: I'm thinking something similar to this: 

https://github.com/microsoft/guidance

Granted, many of the features are similar to xef, but specifically In speaking about the templating language triggering the appropriate agents and structured completions here.

It's something I think we can expand in follo-on features,  certainly.

core/src/commonMain/kotlin/com/xebia/functional/xef/prompt/expressions/Expression.kt
```diff
@@ -0,0 +1,98 @@
...
+      Message(
+        role = Role.USER,
+        content =
+          ExpertSystem(
+              system = "You are an expert in replacing variables in templates",
+              query =
+                """
+          |I want to replace the following variables in the following template:
+          |<template>
+          |${messages.joinToString("\n") { it.content }}
```
javipacheco, line 55: I think that we are trying to replace values only in the last message, it doesn't make sense to replace placeholders in the previous messages because they are messages in the conversation

Can we send only the last message here? I think that we can reduce the number of tokens here

javipacheco (changes requested):
I would like to know your opinion about my message

core/src/commonMain/kotlin/com/xebia/functional/xef/prompt/expressions/Expression.kt
raulraja: We can use generation at any point. Only generation triggers value replacement. They can also be in an assistant message, so we need to process all messages.
For example:
```kotlin
system { "You are an expert role player creator for CyberPunk 2077 and know how to use tools ${prompt("cyberpunk_tools")}" }
```

raulraja: To say some other way, multiple Messages form a single template, or at least that is what this implementation does.

javipacheco (approved):
I was worried about creating different `Expression`s with the same `conversationId`, because we create several `SYSTEM` messages in the middle of the conversation

After my local tests, this is not really a problem. I have created 2 different plans with the same `conversationId` and got the expected results and did not mix the messages.

Anyway, `Expression` now has no control over the conversations, we are always using the conversation in the `Scope`. Maybe we should provide a way for doing that by passing a `conversationId` as a parameter in `Expression.run`. It's just a suggestion

🚀 LGTM!

PR #299, opened 2023-08-07
Add Server module

Montagon:
The objective of this PR is to add a new module in Xef that will allow the use of the library as an API by exposing different endpoints.

This PR will add the endpoint of the `PromptMessage`, supporting OpenAI. The rest of the endpoints will be added in future PRs.

- [x] Provisioning.
- [x] API Header for setting the provider (OpenAI, GCP...), model and auth token.
- [x] Endpoint for `chat/completion`. With the same schema as OpenAI.
- [ ] ~Add docker support for running the server.~ It will be added in a new PR.


raulraja (approved):

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Requests.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+
+
+@Serializable
+@Resource("/chat/completions")
+data class PromptMessageRequest(
+    val model: String,
+    val messages: List<ChatMessage>,
+    val temperature: Double = 1.0,
```
javipacheco, line 29: Are we sure that we want `temperature` =  `1.0` by default?

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -0,0 +1,83 @@
...
+    authenticate("auth-bearer") {
+        post("/chat/completions") {
+            val model: Chat = call.request.headers["xef-model"]?.let {
+                it.toOpenAIModel()
+            } ?: DEFAULT_CHAT
+            val token = call.principal<UserIdPrincipal>()?.name ?: throw IllegalArgumentException("No token found")
+            val scope = CoreAIScope(OpenAIEmbeddings(OpenAI(token).GPT_3_5_TURBO_16K))
+            val data = call.receive<ChatCompletionRequest>().toCore()
+            response<String, Throwable> {
+                model.promptMessage(
```
javipacheco, line 33: Remember, when we are using `model.promptMessage` instead of `ai {}` block, we don't have conversations. 

I think that now is our best option, but we have to keep it in mind in case we want to have conversations in the future

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Requests.kt
```diff
@@ -0,0 +1,38 @@
+package com.xebia.functional.xef.server.http.routes
+
+import io.ktor.resources.*
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
```
raulraja, line 7: Are these models in use? If we are using the ones from the openai client lib we can get rid of these.

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Requests.kt
```diff
@@ -0,0 +1,38 @@
...
+        public val User: ChatRole = ChatRole("user")
+        public val Assistant: ChatRole = ChatRole("assistant")
+        public val Function: ChatRole = ChatRole("function")
+    }
+}
+
+
+@Serializable
+@Resource("/chat/completions")
+data class PromptMessageRequest(
```
javipacheco, line 26: We are not using `PromptMessageRequest` in our code. Are we planning to use it in the future?

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -0,0 +1,83 @@
...
+import io.ktor.server.response.*
+import io.ktor.server.routing.*
+import io.ktor.util.pipeline.*
+import com.xebia.functional.xef.llm.models.chat.ChatCompletionRequest as XefChatCompletionRequest
+
+@OptIn(BetaOpenAI::class)
+fun Routing.routes() {
+    authenticate("auth-bearer") {
+        post("/chat/completions") {
+            val model: Chat = call.request.headers["xef-model"]?.let {
```
javipacheco, line 26:
```suggestion
val model: Chat = call.request.headers["xef-model"]?.toOpenAIModel() ?: DEFAULT_CHAT
```

javipacheco (dismissed):
Just a minor suggestions 🚀

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
raulraja: It automatically adds the `conversationId` if you pass it through as argument or is one of the methods that takes the VectorStore

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Requests.kt
Montagon: I think the three models are not needed, you are right. Removed, thanks!

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
Montagon: Modified, thanks!

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
Montagon: As this is based on the OpenAI API specification, it doesn't have a `conversationId`, instead they have the `user` that is making the call. And it's a `String`, so if you want we can use it as the `conversationId`. But I'm not sure.

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Requests.kt
Montagon: Yes, it's better not to have this value. But now this class doesn't exist, so problem solved 🤓

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
raulraja: We can address it in a different PR, the conversation ID has to be persisted and looked up on each call somehow optionally

raulraja (approved):
Thank you @Montagon , excited about the upcoming Xef server!

PR #305, opened 2023-08-08
New strategy for collecting messages and improvements to ReAct agent and message passing

raulraja:
This PR contains a new strategy for collecting messages by the maximum tokens for the model.

Basically, when we have to send messages to OpenAI we can get the messages from 3 places:

- `New Messages`: they are the messages that we are sending. These messages **always** should be sent if you  don't have tokens for this, we will receive an `AIError` exception
- `History`: they are the previous messages in the conversation
- `Context`: they are documents in the `VectorStore`, for example, if you loaded a PDF

In the new strategy, `fitMessagesByTokens` method,  we are calculating the remaining messages, removing the tokens for the OpenAI response and the tokens for the new messages that are required.

When we have the remaining token, we are using 50% of the token for the history messages and the other 50% for the documents in the `VectorStore`

In future PRs, we can add the percentage for every context in the `PromptConfiguration`


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
```diff
@@ -46,35 +46,23 @@ interface Chat : LLM {
...
+    val request =
       ChatCompletionRequest(
         model = name,
         user = promptConfiguration.user,
-        messages = messages,
+        messages = messagesForRequest,
         n = promptConfiguration.numberOfPredictions,
         temperature = promptConfiguration.temperature,
-        maxTokens = checkTotalLeftChatTokens(),
+        maxTokens = promptConfiguration.minResponseTokens,
```
javipacheco, line 65: I think that the field in the `PromptConfiguration` should be `maxResponseTokens` instead of `minResponseTokens`

raulraja: Let's address it in another PR for prompt improvements unless you want to change it here.

javipacheco: In another PR is good to me

jackcviers-contractor (changes requested):
Most of this is comments, except the fold right.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
```diff
@@ -307,30 +283,62 @@ interface Chat : LLM {
...
+
+    if (messagesTokens >= remainingTokens) {
+      throw AIError.PromptExceedsMaxRemainingTokenLength(messagesTokens, remainingTokens)
+    }
+
+    val remainingTokensForContexts = remainingTokens - messagesTokens
+
+    // TODO we should move this to PromptConfiguration
+    val historyPercent = 50
+    val contextPercent = 50
```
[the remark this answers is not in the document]
raulraja: This distribution should not be hardcoded but instead part of the prompt configuration.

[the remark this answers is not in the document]
javipacheco: I totally agree about 50/50 is not the best strategy, we are thinking about creating a `MessagePolicy` or something like that, with the strategy for selecting the best messages.

Regarding the problem in the [prompt with instructions](https://github.com/xebia-functional/xef/blob/84271af07204fc9ea10a4afeb7d8da3614453342/integrations/sql/src/main/kotlin/com/xebia/functional/xef/sql/SQL.kt#L73-L106) I think this is not a problem. The prompt is never removed, we are removing messages only from the history and in the current implementation in `main` branch we are removing messages too with the `memoryLimit` configuration

In this new strategy, we are fitting the message to the remaining tokens, but the history is always limited

Anyway, I think that we should change some parts of this. I really appreciate your comment

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
```diff
@@ -307,30 +283,62 @@ interface Chat : LLM {
...
+    val historyMessagesWithTokens = history.map { Pair(it, tokensFromMessages(listOf(it))) }
+
+    val totalTokenWithMessages =
+      historyMessagesWithTokens.reversed().fold(Pair(0, emptyList<Message>())) { acc, pair ->
+        if (acc.first + pair.second > maxHistoryTokens) {
+          acc
+        } else {
+          Pair(acc.first + pair.second, acc.second + pair.first)
+        }
+      }
```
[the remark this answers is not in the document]
javipacheco: Good catch!

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
```diff
@@ -136,6 +99,23 @@ interface Chat : LLM {
   ): List<String> =
     promptMessages(Prompt(question), context, conversationId, functions, promptConfiguration)
 
+  @AiDsl
+  suspend fun promptMessages(
```
Montagon, line 103: I see that we are passing a `prompt` in this function. Should the function be renamed to `promptMessage`?

Montagon (approved):

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
javipacheco: I think when these methods were created the plural was related to the return type, not to the first parameter

`promptMessage` returns `String`

`promptMessages` returns `List<String>`

Montagon (approved):
🚀

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/tools/ReActAgent.kt
```diff
@@ -32,172 +37,130 @@ class ReActAgent(
...
-              """
-              .trimMargin(),
-          instructions =
-            listOf(
-              "Provide the final answer to the `input` in a sentence or paragraph",
-            )
-        )
+      messages =
+        listOf(
+          systemMessage { "You are an expert in providing answers" },
```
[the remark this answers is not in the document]
raulraja: Let's address this in a different PR, we need to remove the old ExpertSystem and change many prompts to message style both in Xef and other projects where we were sending everything in one single user message.

For now this works, below it contains additional instructions:

```kotlin
userMessage { "Provide the final answer to the `input` in a sentence or paragraph" },
userMessage { "input: $input" },
assistantMessage {
  "I should create a AgentFinish object with the final answer based on the thoughts and observations"
}
```

PR #308, opened 2023-08-10
Provides lint rules for kotlin jvm compat and fixes issues

jackcviers:
* Provides JvmInline value class rule inline value classes may allow the removal of the `JvmInline` annotation in the future. This would make it impossible to call the class from other jvm languages that are not passing through the kotlin compiler.
* Provides JvmStatic constructor rule for kotlin data classes Kotlin data classes require a final parameter that is generated by the kotlin compiler. This rule requires these classes to have a `JvmStatic` annotated factory method in the companion object of the data class to allow it to be constructed from other JVM languages.
* Applies these rules to all xef projects.


detekt-rules/src/main/kotlin/org/example/detekt/PublicDataClassConstructor.kt
```diff
@@ -0,0 +1,145 @@
...
+												hasJvmStaticConstructor = visitor.hasJvmStaticConstructor
+                    }
+                }
+            }
+            if (!hasJvmStaticConstructor) {
+                report(
+                        CorrectableCodeSmell(
+                                issue,
+                                Entity.from(klass),
+                                "Kotlin data classes must have a JVM static factory method to be compatible with other jvm lanugages.",
```
raulraja, line 51:
```suggestion
                                "Kotlin data classes must have a JVM static factory method to be compatible with other jvm languages.",
```

core/src/commonMain/kotlin/com/xebia/functional/xef/embeddings/Embeddings.kt
```diff
@@ -1,8 +1,13 @@
 package com.xebia.functional.xef.embeddings
 
 import com.xebia.functional.xef.llm.models.embeddings.RequestConfig
+import kotlin.jvm.JvmStatic
 
-data class Embedding(val data: List<Float>)
+data class Embedding(val data: List<Float>) {
+  companion object {
+    @JvmStatic fun apply(data: List<Float>): Embedding = Embedding(data)
```
raulraja, line 8: I thought data classes could be instantiated the same from Java. Why do they need a static `apply`?

jackcviers: See https://github.com/xebia-functional/xef/issues/275#issuecomment-1652423758

In scala, Kotlin data classes report to the compiler a final, non-default, compiler-generated `kotlin.jvm.internal.DefaultConstructorMarker` as one of their arguments. This happens even if you don't have default arugments in the data class constructor, and `@JvmOverloads` does not fix the issue.

PR #310, opened 2023-08-10
Generalizes the checks

fedefernandez:
This PR tries to make the check generic enough so they can be loaded from a configuration file or other storage.


raulraja:
Thank you @fedefernandez !

PR #312, opened 2023-08-10
Search wikipedia tool

Zevleg:
Ticket [Implement search wikipedia in React Agent](https://app.clickup.com/t/85ztrbam3)

This is the implementation of search Wikipedia with ReActAgent. As `WikipediaClient` have 3 functions with diferent params, I created a new `ToolWikipedia` Interface because if I implement `Tool` interface I must to implement the function `invoke` only with one parameter. 

Also I created `ToolMain` interface, where `Tool` and `ToolWikipedia` implements `ToolMain` and it is declared in `ReActAgent` to choose wich `invoke` function run for each tool with when(tool) like this:

```kotlin
val observation: String? =
    when (tool) {
      is ToolWikipedia -> tool.invoke(plan.toolInput, null, null)
      is Tool -> tool.invoke(plan.toolInput)
      else -> null
    }
``` 

Captura de pantalla 2023-08-10 a la(s) 5 52 35 p m


reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/tools/ReActAgent.kt
```diff
@@ -171,8 +171,16 @@ class ReActAgent(
       is AgentAction -> {
         logger.info { "🤔 ${plan.thought}" }
         logger.info { "🛠 ${plan.tool}[${plan.toolInput}]" }
+
+        val tool = tools.find { it.name.equals(plan.tool, ignoreCase = true) }
+
         val observation: String? =
-          tools.find { it.name.equals(plan.tool, ignoreCase = true) }?.invoke(plan.toolInput)
+          when (tool) {
+            is ToolWikipedia -> tool.invoke(plan.toolInput, null, null)
```
raulraja, line 179: I don't think we should alter the hierarchy of tools with `ToolMain` and introduce a hard check here for this tool.
Instead, it would be better suited as three different tools, one for each capability. Since most use cases are natural language input, the helpful tool is only the search tool. The other two are secondary in case the search tool returns relevant results containing page ids or entity names.
Most use cases won't include users explicitly typing the page id or name but instead, some prompt that will derive a search input. Search is the primary use case, and the others are secondary.

With this in mind, I think a design that does not require hardcoding would be composed of 3 tools:
- Search (primary)
- Page Id (secondary) States in its description that it can only be used with valid Wikipedia page ids returned by the primary search tool
- Page Name (secondary) States in its description that it can only be used with valid Wikipedia page names returned by the primary search tool.

In the end, the ReAct agent should be able to use the secondary tools only if it has first performed a Wikipedia search with the primary tool.

raulraja (changes requested):
Thanks, @Zevleg; looks excellent.
We should consider not to hardcode the Wikipedia tool inside the agent checks and instead provide three tools as outlined in my comment. We can't make exceptional cases for each tool, and the agents should work under a single common interface agnostic to the tools.
One of the reasons is that tools files and types may not live in this module but instead on a Wikipedia-specific module down the road.

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/tools/ReActAgent.kt
Zevleg: Done

raulraja (dismissed):
Looks great, thanks @Zevleg !

raulraja (approved):
thank you @Zevleg 👏

PR #315, opened 2023-08-11
GCP Pipeline Jobs

fedefernandez:
API to expose the VertexAI endpoints for custom models (fine-tuning)

### DIFF
--- a/integrations/gcp/src/commonMain/kotlin/com/xebia/functional/xef/gcp/pipelines/GcpPipelinesClient.kt
+++ b/integrations/gcp/src/commonMain/kotlin/com/xebia/functional/xef/gcp/pipelines/GcpPipelinesClient.kt
@@ -6,14 +6,15 @@ import io.ktor.client.HttpClient
 import io.ktor.client.call.body
 import io.ktor.client.plugins.*
 import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
-import io.ktor.client.request.header
-import io.ktor.client.request.get
+import io.ktor.client.request.*
 import io.ktor.client.statement.bodyAsText
 import io.ktor.http.ContentType
 import io.ktor.http.HttpStatusCode
 import io.ktor.http.contentType
 import io.ktor.http.isSuccess
 import io.ktor.serialization.kotlinx.json.json
+import kotlinx.datetime.Instant
+import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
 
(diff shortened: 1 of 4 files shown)


integrations/gcp/src/commonMain/kotlin/com/xebia/functional/xef/gcp/pipelines/GcpPipelinesClient.kt
```diff
@@ -0,0 +1,191 @@
...
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+
+@OptIn(ExperimentalStdlibApi::class)
+class GcpPipelinesClient(
+  private val tuningLocation: String,  // Supported us-central1 or europe-west4
+  private val projectId: String,
+  private val token: String
+) : AutoCloseable, AutoClose by autoClose() {
+  private val http: HttpClient = HttpClient {
```
raulraja, line 27: We seem to repeat this pattern in all clients. Perhaps we should move it to a place with a single configuration with sane defaults for those with the same config.

raulraja (commented):
Looking great @fedefernandez !!

integrations/gcp/src/commonMain/kotlin/com/xebia/functional/xef/gcp/pipelines/GcpPipelinesClient.kt
nomisRev: This class doesn't really need to implement `AutoClose by autoClose()`, but if we do we can do something like this.

```kotlin
  fun AutoClose.http(): HttpClient =
  autoClose(HttpClient {
    install(HttpTimeout) {
      requestTimeoutMillis = 60_000
      connectTimeoutMillis = 60_000
    }
    install(HttpRequestRetry)
    install(ContentNegotiation) {
      json(
        Json {
          encodeDefaults = false
          isLenient = true
          ignoreUnknownKeys = true
        }
      )
    }
  })
```

And then inside all these classes we can just do

```kotlin
private val http: HttpClient = http()
```

This requires updating `AutoClose` to support Ktor's Closeable:

```kotlin
interface AutoClose : AutoCloseable {
  fun <A : AutoCloseable> autoClose(autoCloseable: A): A
  fun <A : Closeable> autoClose(autoCloseable: A): A
}
```

nomisRev (commented):
This PR looks pretty good to me so far. Only added a small reply to Raul's comment.

Intex32:
> This PR looks pretty good to me so far. Only added a small reply to Raul's comment.

I talked to the others, Fede is currently on vacation. You can apply your suggestions and finish this pr, as far as I am concerned.

Intex32:
I'm currently on it. @nomisRev

javipacheco (approved):
🚀 LGTM!

PR #316, opened 2023-08-11
Memory limit by tokens

javipacheco:
This PR changes the strategy for getting the messages in the conversation. 

Now, we have a `limit` and we get the last messages for this limit. In this PR we are adding a new field to `Messages`, the `approxTokens`. In this field, we have the approximate token for every message. 

When we call the `memories` function in the `VectorStore`, now we have to pass the number of token that we want to receive for the last messages in the conversation

```kotlin
suspend fun memories(conversationId: ConversationId, limitTokens: Int): List<Memory>
```

This optimises the way to get the message from the `VectorStore` and we are sending only the necessary tokens to OpenAI

This PR changes all of the implementations of `VectorStore`: `LocalVectorStore`, `CombinedVectorStore` , `PGVectorStore` and `Lucene`


core/src/commonMain/kotlin/com/xebia/functional/xef/vectorstores/LocalVectorStore.kt
```diff
@@ -46,9 +46,20 @@ private constructor(private val embeddings: Embeddings, private val state: Atomi
     }
   }
 
-  override suspend fun memories(conversationId: ConversationId, limit: Int): List<Memory> {
+  override suspend fun memories(conversationId: ConversationId, limitTokens: Int): List<Memory> {
     val memories = state.get().orderedMemories[conversationId]
-    return memories?.takeLast(limit).orEmpty().sortedBy { it.timestamp }
+    return memories
+      ?.foldRight(Pair(0, emptyList<Memory>())) { memory, (accTokens, list) ->
```
raulraja, line 52: This seems to be similar in all implementations, can we do it in a single place and call the function?

javipacheco:
@raulraja @Montagon I have tested the `xef-server` on my local machine and the new `approx_tokens` field in `xef_memory` table is created. Everything works as expected

PR #317, opened 2023-08-11
Update OpenAI defaults selection

Montagon:
This PR changes the way we manage the Defaults values for OpenAI.


raulraja (approved):
thanks @Montagon !

PR #319, opened 2023-08-13
Simplify conversations and their JVM integrations

raulraja:
This PR simplifies the way language support is built for Java and Scala by making the `Conversation` methods available in a JVM-specific implementation that uses a `CoroutineScope` with a supervisor to turn the suspend functions into versions that return `CompletableFuture,` and that can be consumed from other langs.
For now, the `Scala` lib and `Java` examples consume those methods in a blocking style assuming LOOM is available.

I'll leave comments on the files in the most relevant parts.


core/build.gradle.kts
```diff
@@ -87,6 +87,11 @@ kotlin {
         implementation(libs.logback)
         implementation(libs.skrape)
         implementation(libs.rss.reader)
+        api(libs.jackson)
```
raulraja, line 90: The Jackson serialization layer is now part of the core-jvm module

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/Conversation.kt
```diff
@@ -9,36 +9,44 @@ import com.xebia.functional.xef.llm.models.images.ImagesGenerationResponse
...
- * dependencies required to run [AI] values, and provides convenient syntax for writing [AI] based
- * programs.
- */
-class Conversation
-@JvmOverloads
-constructor(
-  val store: VectorStore,
-  val conversationId: ConversationId? = ConversationId(UUID.generateUUID().toString())
-) : AutoCloseable, AutoClose by autoClose() {
+interface Conversation : AutoClose, AutoCloseable {
```
raulraja, line 20: This has changed to an interface to have the implementation be platform specific via `expect/actual` multiplatform declarations. This allows us to project only the `CompletableFuture` based methods over the JVM implementation.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/Conversation.kt
```diff
@@ -9,36 +9,44 @@ import com.xebia.functional.xef.llm.models.images.ImagesGenerationResponse
...
+
+  val store: VectorStore
+
+  val conversationId: ConversationId?
+
+  val conversation: Conversation
 
   @AiDsl
+  @JvmSynthetic
```
raulraja, line 29: Every `suspend` function in the library that is supposed to be a public API translatable to `@CompletableFuture` should be marked as `@JvmSynthetic` so that the methods become hidden from Java and Scala.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/Conversation.kt
```diff
@@ -9,36 +9,44 @@ import com.xebia.functional.xef.llm.models.images.ImagesGenerationResponse
...
   suspend fun addContext(vararg docs: String) {
     store.addTexts(docs.toList())
   }
 
+  fun CoroutineScope.addContextAsync(vararg docs: String): Deferred<Unit> = async {
```
raulraja, line 34: We use a `CoroutineScope` around functions we want to publish as `Deferred` or a `CoroutineScope(SupervisorJob)` to invoke these or wrap suspend functions directly.

core/src/commonMain/kotlin/com/xebia/functional/xef/auto/Conversation.kt
```diff
@@ -101,10 +158,37 @@ constructor(
...
+  ): Deferred<ImagesGenerationResponse> = async {
+    images.images(prompt, store, numberImages, size, promptConfiguration)
+  }
+
+  companion object {
+
+    operator fun invoke(
+      store: VectorStore,
+      conversationId: ConversationId? = ConversationId(UUID.generateUUID().toString())
+    ): PlatformConversation = PlatformConversation.create(store, conversationId)
```
raulraja, line 185: Having `PlatformConversation` as `expect` allows us to define factories and constructors for it in the common module.

java/src/main/java/com/xebia/functional/xef/java/auto/AIScope.java
raulraja, line 1: The AIScope wrapping is no longer needed in Java since the `JVMConversation` includes the API for futures

reasoning/src/jvmMain/kotlin/com/xebia/functional/xef/reasoning/pdf/ReadPDFFromUrl.kt
```diff
@@ -28,6 +33,12 @@ constructor(
 
   override val description: String = "Reads the content of a PDF as String from a URL"
 
+  private val coroutineScope = CoroutineScope(SupervisorJob())
```
raulraja, line 36: In some cases is convenient in services like this one to declare a `coroutineScope` that allows turning `suspend` APIs and offering a version for `CompletableFuture`. In this case, we do it like this because the PDF functionality is only available in the JVM at the moment of writing this.

scala/src/main/scala/com/xebia/functional/xef/scala/auto/package.scala
```diff
@@ -1,101 +1,97 @@
...
+)(using
+    conversation: ScalaConversation
 ): List[String] =
-  LoomAdapter
-    .apply[java.util.List[String]](count =>
-      resource match
-        case url: String => Loader.pdf(url, splitter.core, count)
-        case file: File => Loader.pdf(file, splitter.core, count)
-    ).asScala.toList
+  conversation.promptMessages(chat, question, functions.asJava, promptConfiguration).join().asScala.toList
```
raulraja, line 51: We assume LOOM for now, and we block here, in practice scala users may want to consume this blocking, with CompletableFuture and with an IO

scala/src/main/scala/com/xebia/functional/xef/scala/auto/package.scala
```diff
@@ -1,101 +1,97 @@
...
+    // TODO change to fs2 or similar
+    def onSubscribe(s: Subscription): Unit = s.request(Long.MaxValue)
+
+    def onNext(t: String): Unit = queue.add(t); ()
+
+    def onError(t: Throwable): Unit = throw t
+
+    def onComplete(): Unit = ()
+  })
+  LazyList.continually(queue.take)
```
raulraja, line 73: We should probably consider replacing this implementation or providing an alternative with something like fs2

javipacheco (approved):
After the conversation this morning with Raul, I understand the importance of this

Really good job! LGTM! 🚀

PR #320, opened 2023-08-13
Diff summary tool and create PR Description example

raulraja:
This PR introduces changes to the following files:

- `core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt`
- `examples/kotlin/src/main/kotlin/com/xebia/functional/xef/auto/reasoning/CodeExample.kt`
- `examples/kotlin/src/main/kotlin/com/xebia/functional/xef/auto/reasoning/CreatePRDescription.kt`
- `reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/code/Code.kt`
- `reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/code/DiffSummary.kt`

The changes include the following:

- Added a new function `promptMessage` to the `Chat` interface in `Chat.kt`.
- Added a new function `main` to `CodeExample.kt` that demonstrates the usage of the `Code` class.
- Added a new file `CreatePRDescription.kt` that contains the implementation of the `CreatePRDescription` function.
- Updated the `Code` class in `Code.kt` to include the `diffSummaryFromUrl` tool.
- Added a new file `DiffSummary.kt` that contains the implementation of the `DiffSummary` tool.

These changes aim to improve the functionality and usability of the codebase. The `DiffSummary` tool allows users to summarize code diffs given a diff URL and create a PR description. This tool can be used to automate the process of creating descriptive and informative PRs.

Please review the changes and provide your feedback.

---

The description above has been generated with the example:
```kotlin
package com.xebia.functional.xef.auto.reasoning

import com.xebia.functional.xef.llm.models.chat.Message
import com.xebia.functional.xef.reasoning.code.Code
import com.xebia.functional.xef.reasoning.tools.ReActAgent

suspend fun main() {
  OpenAI.conversation {
    val code =
      Code(
        model = OpenAI().DEFAULT_CHAT,
        serialization = OpenAI().DEFAULT_SERIALIZATION,
        scope = this
      )

    val agent =
      ReActAgent(
        model = OpenAI().DEFAULT_SERIALIZATION,
        scope = this,
        tools = listOf(code.diffSummaryFromUrl)
      )
    val prDescription =
      agent.run(
        listOf(
          Message.userMessage {
            "Create a PR description for https://github.com/xebia-functional/xef/compare/simplify-future-integration-and-conversations...diff-summary-tool.diff"
          }
        )
      )
    println(prDescription)
  }
}
```

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
@@ -103,6 +103,15 @@ interface Chat : LLM {
   }
 
   @AiDsl
+  suspend fun promptMessage(
+    messages: List<Message>,
+    scope: Conversation,
+    promptConfiguration: PromptConfiguration = PromptConfiguration.DEFAULTS
+  ): String =
+    promptMessages(messages, scope, emptyList(), promptConfiguration).firstOrNull()
+      ?: throw AIError.NoResponse()
+
+  @AiDsl
   suspend fun promptMessages(
     messages: List<Message>,
     scope: Conversation,
--- a/reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/code/DiffSummary.kt
+++ b/reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/code/DiffSummary.kt
@@ -1,0 +1,130 @@
+package com.xebia.functional.xef.reasoning.code
+
+import com.xebia.functional.xef.auto.AutoClose
+import com.xebia.functional.xef.auto.Conversation
+import com.xebia.functional.xef.auto.autoClose
+import com.xebia.functional.xef.llm.Chat
+import com.xebia.functional.xef.llm.ChatWithFunctions
+import com.xebia.functional.xef.llm.models.chat.Message
+import com.xebia.functional.xef.prompt.Prompt
+import com.xebia.functional.xef.reasoning.text.summarize.Summarize
+import com.xebia.functional.xef.reasoning.text.summarize.SummaryLength
+import com.xebia.functional.xef.reasoning.tools.Tool
+import io.github.oshai.kotlinlogging.KLogger
+import io.github.oshai.kotlinlogging.KotlinLogging
+import io.ktor.client.*
+import io.ktor.client.request.*
+import io.ktor.client.statement.*
+import io.ktor.http.*
+import kotlinx.serialization.Serializable
+
+class DiffSummary(
+  private val serialization: ChatWithFunctions,
+  private val chat: Chat,
+  private val scope: Conversation
+) : Tool, AutoClose by autoClose(), AutoCloseable {
+
+  private val client = HttpClient {}
+
+  private val summarize =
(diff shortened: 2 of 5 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
```diff
@@ -102,6 +102,15 @@ interface Chat : LLM {
     return promptMessages(prompt.toMessages(), scope, functions, promptConfiguration)
   }
 
+  @AiDsl
+  suspend fun promptMessage(
```
javipacheco, line 106: Should we use this method for the `promptMessage(question: String, ....)` function in line 78?

Something like this

```kotlin
  @AiDsl
  suspend fun promptMessage(
    question: String,
    scope: Conversation,
  ): String =
    promptMessage(question.toMessages(), scope, promptConfiguration)
```

And, maybe we should also create  `fun promptMessage(prompt: Propmt,...)` and we will have the 3 implementations in the same way that the plural functions `promptMessages`

javipacheco (changes requested):
Just a minor comments

reasoning/src/commonMain/kotlin/com/xebia/functional/xef/reasoning/code/DiffSummary.kt
```diff
@@ -0,0 +1,130 @@
...
+
+  @Serializable data class ExtractedUrl(val url: String)
+
+  override suspend fun invoke(input: String): String {
+    val extracted: ExtractedUrl =
+      serialization.prompt(
+        prompt =
+          Prompt(
+            """|
+        |Please provide the url that you want to read a PDF from
```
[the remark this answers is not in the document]
raulraja: bad copy pasta, fixed

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
raulraja: To do this we need a form of template separator in the string such as:
```txt
system:
  This is valid yaml parseable as a system message
user: 
  This goes to a user message  
```
Otherwise, it all shows up as a user message, which is the same as we have now.
I think this makes sense to implement when we have Prompt persistence and loading prompts by ID.

A user wants a prompt to persist and edit it as a whole but the structure should remain so we can always go back to the list of messages.

raulraja: I also added `promptMessage` which returns the single String response but takes a `List<Message>`

raulraja: I think I missunderstood your original question and gave you instead an answer on how I was thinking on persisting templates as String, yes we could use the promptMessage method but since that one currently returns a String and not a serialized `A` which is what we need here with precision to clean up and parse the previous input. 
We could introduce a version of serializable prompts that take a List of messages and refactor wherever makes sense.

javipacheco (approved):
🚀  LGTM!

PR #326, opened 2023-08-14
New PromptBuilder returning list of messages

javipacheco:
This PR changes the way to make prompts. This is the new way:

```kotlin
      val messages = buildPrompt {
        +system("Test System")
        +user("Test Query")
        +assistant("Test Assistant")
      }
```

`buildPrompt` returns a `List<Message>` that you can pass to `prompt` functions in the conversation scope.

This PR removes:

- `ExpertSystem`
- Old `Message` model that we are not using in `xef.prompt` package
- The suspend functions in the Message companion object


core/src/commonMain/kotlin/com/xebia/functional/xef/prompt/PromptBuilder.kt
```diff
@@ -0,0 +1,25 @@
+package com.xebia.functional.xef.prompt
+
+import com.xebia.functional.xef.llm.models.chat.Message
+import com.xebia.functional.xef.llm.models.chat.Role
+
+open class PromptBuilderBuilder {
```
raulraja, line 6: PromptBuilderBuilder?

raulraja (approved):
Looks great @javipacheco

Montagon (approved):

PR #329, opened 2023-08-17
Prevents jvm incompat dataclass constructors with value class params

jackcviers:
* To prevent issues like #275, dataclasses must not use parameters
that are value classes without a `JvmStatic` factory method in their
companion object.


detekt-rules/src/main/kotlin/org/example/detekt/PublicDataClassConstructorWithValueParameters.kt
```diff
@@ -0,0 +1,98 @@
...
+              )
+            }
+          }
+        }
+        if (!hasJvmStaticConstructor) {
+          report(
+            CorrectableCodeSmell(
+              issue,
+              Entity.from(ktClass),
+              "Kotlin data classes must have a JVM static factory method to be compatible with other jvm languages.",
```
raulraja, line 88: I thought we discovered this was not the case if the constructor was properly annotated as `@JvmOverloads constructor`. Why do we need to enforce a static factory method?

detekt-rules/src/test/kotlin/org/example/detekt/PublicDataClassConstructorWithValueParametersTest.kt
```diff
@@ -0,0 +1,106 @@
...
+internal class PublicDataClassConstructorWithValueParametersTest(
+  private val env: KotlinCoreEnvironment
+) {
+  @Test
+  fun `reports data classes that have value class parameters without companion static constructor`() {
+    val code =
+      """
+        import kotlin.jvm.JvmInline
+        @JvmInline value class  Bar(val value: String)
+        data class Foo(val bar: Bar)
```
raulraja, line 20: Is the problem here the same as Rule 1?, that we are using an `inline value class` as a property in a data class?
Do we need need to test that data classes have static methods? I think instead of testing that if there is a static factory we should just disallow value classes altogether and forbid their usage.

raulraja (changes requested):
Thank you, Jack!

After looking over the files, it seems at first sight all issues here are related to the use of the `value` class.
If that is the case we need a rule to disallow using `value` classes.
I think the rule for static companion factories is unnecessary, and in main, we now have instances of multiple cases where the `Prompt` class, a data class, is called from Java to any of its constructors.

Can we have a single rule to disallow `value` instead of 2 rules?

detekt-rules/src/main/kotlin/org/example/detekt/PublicDataClassConstructorWithValueParameters.kt
jackcviers-contractor:
```suggestion
              "Kotlin data classes containing parameters that are value classes must have a JVM static factory method to be compatible with other jvm languages.",
```

jackcviers-contractor (changes requested):

raulraja:
We can make those changes to `RequestConfig`. We are currently ongoing heavy API changes to simplify the library.
The reality is that having used value classes has introduced more issues than benefits, and the code we have to add to work around them is higher than the benefit of saving an allocation.

raulraja (dismissed):

raulraja (approved):
thank you Jack!

PR #330, opened 2023-08-17
Simplify access to functions and prompts

raulraja:
This PR simplifies how we thread functions and makes Prompts aware of the functions.
In the following PR, I plan on moving out of the `Chat` and `ChatWithFunctions` interfaces, the logic related to counting tokens for messages, and similar inside the Prompt area where all the information sent now that it includes functions contained in a single place.


Montagon (approved):
🚀

PR #333, opened 2023-08-18
Add @Description annotations for Java and Scala

raulraja:
Fixes annotation so they work in Java and Scala.

1. expect annotation in common
2. Java file with annotation in jvmMain
3. The Java annotation becomes the actual impl for the expect
4. Scala declares a static annotation and implements the Jvm Annotation
5. Jackson spits out the Json schema through a new module that reads the annotations that are in the hierarchy of Description
6. Scala uses a macro to extract the static annotations when building the serial descriptor since Mirror.Of lacks support for annotations


core/src/jvmMain/kotlin/com/xebia/functional/xef/auto/Description.jvm.kt
```diff
@@ -0,0 +1,8 @@
+package com.xebia.functional.xef.auto
+
+import com.xebia.functional.xef.auto.jvm.Description
+
+/**
+ * Description of a property to enhance the LLM prompt chances to get the data in the right format
+ */
+actual typealias Description = Description
```
jackcviers, line 8: That's how you do this! I was looking for this!

jackcviers (approved):
Thank you! LGTM

Montagon (approved):
Amazing! 🚀

PR #334, opened 2023-08-18
Prompt messages with serializable objects

javipacheco:
This PR removes the `fun prompt(input:A) : B` method and we are adding the input serialization to the PromptBuilder.

The previous approach has different problems:

- Since the `Prompt` is a `@Serializable` sometimes we were sending the prompt data to this method
- We are creating always only one `user` message, and we should be able to create `assistant` and `system` messages

Now your prompt could be something like that:

```kotlin
@Serializable data class Question(val question: String)
val question = Question("What is your name?")

Prompt {
  +system("System message")
  +user("User message") 
  +user(question)
}
```

And you can add several serialized messages to your prompt


raulraja (approved):
I discussed this offline with @javipacheco and @Montagon , this looks great but we need to address also the Java and Scala api

raulraja (approved):
Looks great @javipacheco !

core/src/commonMain/kotlin/com/xebia/functional/xef/prompt/templates/templates.kt
```diff
@@ -5,13 +5,22 @@ import com.xebia.functional.xef.llm.models.chat.Role
...
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.serializer
 
 fun system(context: String): Message = context.message(Role.SYSTEM)
 
 fun assistant(context: String): Message = context.message(Role.ASSISTANT)
 
 fun user(context: String): Message = context.message(Role.USER)
 
+inline fun <reified A> system(data: A): Message = system(Json.encodeToString(serializer(), data))
```
jackcviers, line 17: Do you happen to know if this can be called from Java or Scala?

I feel like they should be,  but haven't checked. 

If you can confirm that,  just mark this as resolved.

jackcviers (changes requested):
Only see a single compatibility question. We could probably define the unary + in xef-scala, but not for xef-java of course. add<X> is good enough for now.

If you can confirm the single comment works across java/scala boundaries,  then ping me an Ill approve.

core/src/commonMain/kotlin/com/xebia/functional/xef/prompt/templates/templates.kt
raulraja: They can't be called from other JVM languages but we also expose a version that is not `inline` `reified` that takes in a serializer and in the JVM impl we can also take in a .class

raulraja: The `unaryPlus` and other methods meant to be used by kotlin are `@JvmSynthetic,` and they don't show up in scala and Java autocompletion or public members.

PR #338, opened 2023-08-21
Using Jackson for adding objects to PromptBuilder

javipacheco:
This PR adds the possibility of generating object messages using `Jackson` from Java and Scala


examples/java/src/main/java/com/xebia/functional/xef/java/auto/jdk8/Animals.java
```diff
@@ -25,7 +26,7 @@ public CompletableFuture<Invention> groundbreakingInvention() {
     }
 
     public CompletableFuture<Story> story(Animal animal, Invention invention) {
-        PromptBuilder builder = new PromptBuilder()
+        PromptBuilder builder = new JvmPromptBuilder()
```
raulraja, line 29: After seeing how this has been laid out, I wonder if we can make the `PromptBuilder` the `expect/actual`. This would allow us to do `new PromptBuilder()`. Otherwise, we are leaking our implementation details to the use of the library, and it forces users to learn that there are different types that they need to use depending on the language.

javipacheco: For that, I have to create all of the abstract methods in `PromptBuilder`, even the `unaryPlus` one, and implement them in every target. I'm not sure if we want to do that

Something like:

```kotlin
expect interface PromptBuilder {
  val items: MutableList<Message>

  fun preprocess(elements: List<Message>): List<Message>

  fun build(): Prompt

  @JvmSynthetic
  operator fun Prompt.unaryPlus() 

  fun addMessage(message: Message): PromptBuilder

  ... all of methods here ...

  companion object {

    fun create(): PromptBuilder
  }
}
```

raulraja (approved):

PR #342, opened 2023-08-22
Function Call fixed

javipacheco:
We are sending empty function calls when it's not necessary and the OpenAI Endpoint complains about it. This PR fixes that and I have created several tests for this scenario


raulraja (approved):
thanks @javipacheco !

core/src/commonTest/kotlin/com/xebia/functional/xef/conversation/ConversationSpec.kt
```diff
@@ -113,4 +118,61 @@ class ConversationSpec :
 
       messagesSizePlusMessageResponse shouldBe memories.size
     }
+
+    "functionCall should be null when the model doesn't support functions" {
```
Montagon, line 122: Don't know if this case is possible:
`model doesn't support functions but the prompt contains a function`.

Just a doubt, great job!!

Montagon (approved):
🚀

PR #345, opened 2023-08-23
Fixed reified String in prompt function

javipacheco:


raulraja (approved):
thanks, @javipacheco for the fix and @Intex32, for finding the issue!

PR #346, opened 2023-08-23
Server support stream

Montagon:


server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -5,65 +5,124 @@ import com.aallam.openai.api.chat.ChatCompletionRequest
...
+            if (!data.containsKey("model")) {
+                call.respondText("No model found", status = HttpStatusCode.BadRequest)
+                return@post
+            }
+            val model: OpenAIModel = data["model"]?.jsonPrimitive?.content?.toOpenAIModel(token) ?: run {
+                call.respondText("No model found", status = HttpStatusCode.BadRequest)
+                return@post
+            }
+
+            println(darta)
```
raulraja, line 70: Probably this log statement can be removed

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -5,65 +5,124 @@ import com.aallam.openai.api.chat.ChatCompletionRequest
...
+                    headers {
+                        bearerAuth(token)
+                    }
+                    contentType(ContentType.Application.Json)
+                    method = HttpMethod.Post
+                    setBody(darta)
+                }
+                call.respond(response.body<String>())
+            } else {
+                runBlocking {
```
raulraja, line 85: why do we need to call `runBlocking` here? 

I think we should be able to get the response straight from OpenAI and use `call.respondOutputStream(openAIResponse.outputStream)` or a similar API where we just connect the two output streams.

We are manually unpacking the bytes and buffering them with a different flush and buffer policy than the one OpenAI is using, so we may not stream back the chunks like they do if we do it like this.

raulraja (dismissed):
We could test this change live by creating our `xef-client`, which uses the `OpenAI` Models we have in Xef but instead uses our server URL. Sending a streaming request like the `SpaceCraft` example that streams the response will show if this integration is wired the same as Open AIs. Let's discuss it tomorrow to see how we can approach it and if we can already write some tests for it.

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -5,65 +5,124 @@ import com.aallam.openai.api.chat.ChatCompletionRequest
...
+    client: HttpClient,
+    persistenceService: PersistenceService
+) {
+    val openAiUrl = "https://api.openai.com/v1"
+
+    authenticate("auth-bearer") {
+        post("/chat/completions") {
+            val provider: Provider = call.getProvider()
+            val token = call.getToken()
+            val scope = Conversation(persistenceService.getVectorStore(provider, token))
```
javipacheco, line 58: We are not using Xef conversations here, should we remove this?

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -5,65 +5,124 @@ import com.aallam.openai.api.chat.ChatCompletionRequest
...
+            val darta = call.receive<String>()
+            val data = Json.decodeFromString<JsonObject>(darta)// call.receive<JsonObject>()
+            if (!data.containsKey("model")) {
+                call.respondText("No model found", status = HttpStatusCode.BadRequest)
+                return@post
+            }
+            val model: OpenAIModel = data["model"]?.jsonPrimitive?.content?.toOpenAIModel(token) ?: run {
```
javipacheco, line 65: `model` is not necessary in the call. I think that we should remove this

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -5,65 +5,124 @@ import com.aallam.openai.api.chat.ChatCompletionRequest
...
-      }
+fun Routing.routes(
+    client: HttpClient,
+    persistenceService: PersistenceService
+) {
+    val openAiUrl = "https://api.openai.com/v1"
+
+    authenticate("auth-bearer") {
+        post("/chat/completions") {
+            val provider: Provider = call.getProvider()
```
javipacheco, line 56: At the moment this endpoint is only for OpenAI, I think that we don't need to send the `xef-provider` header in the call

javipacheco (changes requested):
Just a minor suggestion. Please, let me know your opinion about that

javipacheco:
@raulraja the last changes added to this PR add the possibility to change the host in order to use the `xef-server` in the call is `SpaceCraftLocal`example

raulraja (approved):
thank you @Montagon @javipacheco !

javipacheco (approved):

PR #347, opened 2023-08-24
Removing duplicates examples

javipacheco:
This PR adapts the examples to the new version

- We are removing all of the java8 examples
- We are removing examples with the same feature. For example, we don't need a lot of examples about serialization
- We are adding every example to the corresponding package according to the feature that performs the example


examples/kotlin/src/main/kotlin/com/xebia/functional/xef/conversation/conversations/Animal.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.xebia.functional.xef.prompt.templates.system
+import com.xebia.functional.xef.prompt.templates.user
+import kotlinx.serialization.Serializable
+
+@Serializable data class Animal(val name: String, val habitat: String, val diet: String)
+
+@Serializable
+data class Invention(val name: String, val inventor: String, val year: Int, val purpose: String)
+
+suspend fun main() {
```
javipacheco, line 15: I have changed this example. In the previous version, we directly added information about the animal and invention to the prompt. Now, this example uses this information in the earlier message of the conversation, showing that OpenAi is receiving all the messages in the conversation

raulraja (approved):
Thank you @javipacheco !

PR #351, opened 2023-08-25
Upgrade JDK version to 20

franciscodr:
The `Scala` modules require using the latest versions of Java because they depend on the `loom`  project for structured concurrency. So far, we have been using version 19, but a new STS version (Java 20) was recently released to replace the former one.

### DIFF
--- a/docs/intro/scala.md
+++ b/docs/intro/scala.md
@@ -76,7 +76,7 @@ Set the environment variable `OPENAI_TOKEN=xxx`
 ### Project Loom
 
 The Scala module depends on project [Loom](https://openjdk.org/projects/loom/), 
-so you will need at least Java 19 to use the library. Furthermore, you need to pass
+so you will need at least Java 20 to use the library. Furthermore, you need to pass
 the `--enable-preview` flag.
 
 <details>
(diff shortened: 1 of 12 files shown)


docs/intro/scala.md
```diff
@@ -90,7 +90,7 @@ env OPENAI_TOKEN=<your-token> sbt -J--enable-preview <your-command>
 <details>
 <summary>IntelliJ</summary>
 
-- Set the Java version to at least 19
+- Set the Java version to at least 20
 - Set VM options to `--enable-preview`
```
raulraja, line 94: Do we still need `--enable-preview` if we set the default to 20?

raulraja (approved):
thanks @franciscodr !

docs/intro/scala.md
franciscodr: Good question, let me check it

franciscodr: According to the [Java documentation](https://openjdk.org/jeps/437), the `StructuredTaskScope` API still resides in the incubator module, so the `--enable-preview` flag is still required.

PR #372, opened 2023-08-31
xef-dashboard basic API

calvellido:
## Description

This PR adds the needed code to be able to connect with `xef-server` API and communicate with it though a generic form. Currently, it uses the `chat/completions` endpoint. In order to use this in the best way, the request body and response have been types following the OpenAI documentation:

https://github.com/openai/openai-node/

To be able to properly exchange data with the API, the changes also add a _Settings_ page, where the user can save its `xef-server`/OpenAI API key to communicate with the server.

It also modifies Ktor config to allow CORS auth, as it previously wouldn't work through the combination of CORS and the `Authorization` header.

## Preview

imagen
imagen


javipacheco (changes requested):
I have been trying the code locally and I have a problem

When I create the website using `npm run dev` the page works as expected, but if I create the website using `./gradlew web` when I send a message I always receive this message `Chat completions request couldn't be completed, is the API key set?`. I'm saving the API key in the settings

On the other hand, if I reload the page in the browser I have to save the API Key again. I understand what the problem is, but should we store the key somewhere?

calvellido:
> I have been trying the code locally and I have a problem
> 
> When I create the website using `npm run dev` the page works as expected, but if I create the website using `./gradlew web` when I send a message I always receive this message `Chat completions request couldn't be completed, is the API key set?`. I'm saving the API key in the settings
> 
> On the other hand, if I reload the page in the browser I have to save the API Key again. I understand what the problem is, but should we store the key somewhere?

Everything works as expected on my end.

Open the browser's dev console and check the requests, if you are having a 401, a 500… Please also check the server logs to see if the request is reaching the server, and something is going up there, even is the response isn't showing it.

Regarding the API key: of course, ideally, the API key could be saved in the `localStorage` or `sessionStorage` to have it persist in or in between browser sessions. Next PRs could address this.

There's probably room to improve in all components, but the PR scope should be set somewhere, otherwise the code might get cut without being finished. For example, in this one, initially the API key was set per request, but I thought this could be a good opportunity to see the difference between local states and contexts in proper code.

javipacheco (approved):
🚀  LGTM!

I was opening the web from http://0.0.0.0:8080 and this is not working because is blocked by CORS policy.  Finally using http://localhost:8080 is working as expected

PR #374, opened 2023-09-01
Added embeddings endpoint to server

Montagon:
This PR include the endpoint for the embeddings in the same way that OpenAI supports it.


server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -60,36 +60,60 @@ fun Routing.routes(
...
-                        bearerAuth(token)
-                    }
-                    contentType(ContentType.Application.Json)
-                    method = HttpMethod.Post
-                    setBody(context)
-                }
-                call.respond(response.body<String>())
+                client.makeRequest(call, "$openAiUrl/chat/completions", context, token)
             } else {
                 runBlocking {
```
raulraja, line 65: I'm unsure if we can use something different here to `runBlocking`. /cc @nomisRev

raulraja (dismissed):

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
```diff
@@ -60,36 +60,60 @@ fun Routing.routes(
...
+}
+
+private suspend fun HttpClient.makeRequest(
+    call: ApplicationCall,
+    url: String,
+    body: String,
+    token: String)
+{
+    val response = this.request(url) {
+        headers {
```
javipacheco, line 86: This header is the same in `makeRequest` and `makeStreaming`, maybe we can put it in a common place

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
javipacheco, line 65: I think that we don't need `runBlocking` here

javipacheco (changes requested):
In order to finish this endpoint we should be sure that the embeddings are working fine. For that, we are using the `SpaceCraftLocal` example

We should change this example using the local embedding. Something like that

```kotlin
  val localOpenAi = OpenAI(host = "http://localhost:8081/")

  val model = localOpenAi.DEFAULT_SERIALIZATION

  val scope =
    Conversation(LocalVectorStore(OpenAIEmbeddings(localOpenAi.DEFAULT_EMBEDDING)))

....
```

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
nomisRev: I'm not entirely sure of the context here but inside `post` we can call `suspend` code so we shouldn't need `runBlocking` at all 🤔

If we need to run something _async_ and return immediately (POST), then we should use `application.launch` where `application` is the property in `post` referencing to the parent server lifecycle.

Intex32:
@Montagon have you tested streaming? when I tested that earlier, it hang forever. probably something was just wrong with my setup.

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
Montagon: Thanks @nomisRev. I tried to replicate the behaviour of the OpenAI endpoint by using the `runBlocking`. I will try your approach 💪

Montagon:
> @Montagon have you tested streaming? when I tested that earlier, it hang forever. probably something was just wrong with my setup.

In which example did you try? I tried the `SpaceCraftLocal` with the new endpoint and it fails, not sure if related. But I'm checking it. Thanks!

Intex32:
> @Montagon have you tested streaming? when I tested that earlier, it hang forever. probably something was just wrong with my setup.

As I found out, whether the request hangs or not depends on the length of the response from OpenAI:
Prompting to "Write an essay about climate change" does always hang, as an essay is a long text.
Short prompts do work as expected.
Now, the threshold appears to be around 80,000 chars (or bytes not sure) - mas o menos.

I haven't figured out why this is.

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/Routes.kt
Montagon: I've just moved the `HttpRequestBuilder` to a function and used it in both cases.

Intex32 (approved):
I think we're set. Venga!

My issue with long prompts is going to be targeted in a different PR.

raulraja (approved):

javipacheco (approved):
🚀

PR #375, opened 2023-09-01
Refactor API service in its own function

calvellido:
## Description

This PR just refactor the `chatCompletions` API functions into its own function, for a better code organization, reusage, and readability

It also tries to show actual errors that may come from the `xef-server` API, even if they incorrectly respond with `200` codes.


Montagon (approved):
🚀

PR #380, opened 2023-09-04
OpenAIEmbeddings to core module

javipacheco:
`OpenAIEmbeddings` shouldn't be part of `OpenAI` module because it doesn't depend on any `openai` stuff. This PR is moving this class to the core module. I'm refactoring the names of some classes:

- `OpenAIEmbeddings` now is `EmbeddingsProvider` on `core` module
- `com.xebia.functional.xef.llm.Embeddings` now is `LLMEmbeddings` in order to avoid conflicts with `com.xebia.functional.xef.embeddings.Embeddings`
- `GcpEmbeddings` was removed. We are using `EmbeddingsProvider` on `GCP`


Montagon (approved):

Intex32:
Considering the scope of the issue, LGTM.
Possible renames for `EmbeddingsProvider` I can think of are EmbeddingsStrategy and EmbeddingsService.

For another issue, I suggest making the `embedQuery` function an extension function that calls `embedDocuments`, as it is just a special case of the latter until now.
Additionally, the chunkSize should be passed in the constructor of `EmbeddingsProvider` and not be part of the `Embeddings` interface function signature, as the chunk size is specific to this implementation.

These are just my humble thoughts.

Intex32 (dismissed):

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Embeddings.kt
```diff
@@ -1,8 +1,28 @@
...
 
 interface Embeddings : LLM {
   suspend fun createEmbeddings(request: EmbeddingRequest): EmbeddingResult
+
+  suspend fun embedDocuments(
+    texts: List<String>,
+    requestConfig: RequestConfig,
+    chunkSize: Int?
+  ): List<Embedding> =
+    if (texts.isEmpty()) emptyList()
```
raulraja, line 17: I think this way of extracting embeddings which we had since the beginning it's not effitient.
Here we split a chunk of text into the `chunkSize` or `400` and send N request to the open AI embeddings endpoint.

In reality the embeddings endpoint by default looks like this:
- cl100k_base encoding with 8191 max input tokens.

I think it'd be more efficient to split the text based on the model's `max input tokens` and then send fewer and larger requests close to the limit.
Something like:
`chunked(modelType.maxTokens - 100).parMap...`

raulraja (approved):
Looks great @javipacheco , thanks!

In this or another PR, we must address the number and size of messages for embedding requests. We have been defaulting to `400` or configured chunk size, but we know the max content length of the model, which would allow us to send fewer and lengthier requests.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Embeddings.kt
javipacheco: This is really interesting. We should address this in a separate PR 👍

PR #381, opened 2023-09-04
fix incorrect impl of flow (streaming gen.) for local LLMs

Intex32:
When playing with local models, I noticed that the generation (returning a flow) is not printed in real time to the console. After investigation, I figured out this was because of the `coroutineScope` in the `Gpt4All` file that waits until all child jobs are done.

I am not very fluent with the `Flow` API, still I believe that I improved the code rather than worsen. Feedback is appreciated. Also consider my comments.


gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/GPT4All.kt
```diff
@@ -117,51 +111,52 @@ interface GPT4All : AutoCloseable, Chat, Completion {
...
-            val flow = channel.consumeAsFlow().map { toChunk(it) }
+        return channel
+          .consumeAsFlow()
+          .map(::toChunk)
+          .onStart {
+            System.setOut(printStream) // Set the standard output to the print stream
 
-            launch(Dispatchers.IO) {
-              System.setOut(printStream) // Set the standard output to the print stream
+            thread(isDaemon = true) { // generate in background and emit values to flow
```
Intex32, line 151: Introducing a coroutine scope inside this function was the original issue. The only solution I saw is creating a thread. Any better suggestions?
I tried to minimize devastation by spawning it inside the `onStart` event.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
```diff
@@ -37,19 +34,14 @@ interface Chat : LLM {
...
-        it.choices.forEach { choice ->
-          val text = choice.delta?.content ?: ""
-          buffer.append(text)
-        }
-      }
-      .onCompletion {
-        val message = assistant(buffer.toString())
+      .map { it.choices.mapNotNull { it.delta?.content }.reduce(String::plus) }
+      .onEach { emit(it) }
+      .fold("", String::plus)
```
Intex32, line 40: Is there a better way to achieve the same as the `onEach` and `fold`? Currently, each item is emitted to another outer flow.
This is required because we have to `fold` the flow and get the full generated string in order to add it to the memory. On the other hand, we want to pass each item forward to the caller of the function so that they can do whatever the want to do with it.

Montagon (dismissed):

Intex32:
@nomisRev Could you perhaps have a look? in particular in respect to my comments

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
nomisRev: It kind-of depends on the desired behavior of the resulting `Flow`.
`onEach` will back-pressure the `fold`, so the resulting `Flow` can still back-pressure the `fold` which is desired.

`onEach` is the same of `map { f(it); it }`. 

My concern is that `fold` will take everything in memory, but I guess that's inherit to `MemoryManagement.addMemoriesAfterStream`. Just going by the name it sounds like it puts everything in memory anyway.

gpt4all-kotlin/src/jvmMain/kotlin/com/xebia/functional/gpt4all/GPT4All.kt
nomisRev: > Introducing a coroutine scope inside this function was the original issue.

Can you elaborate? These two should be equivalent, except that the `launch` was probably not properly scoped to the `Flow`. I think this could work.

```kotlin
merge(
  suspend {
    generateCompletion(prompt, config, request.streamToStandardOut)
    channel.close()
  }.asFlow().flowOn(Dispatchers.IO),
  channel
    .consumeAsFlow()
    ...
)
```

`merge` runs two `Flow` in parallel. Here we use a `Flow` that emits no elements to execute the side-effect in parallel.

Intex32: I tried to incorporate your points in the latest commit.

Clarifying: Previously, a coroutineScope was used to `lauch` the generation of next tokens _supposedly_ in the background. As the coroutineScope call didn't finish until the launch job finishes, the function didn't return and thus there was no streaming effect.

You suggested merging two flows and make flow generating text run on IO. Better than creating thread! That appears to be working. Thanks. Instead of converting a suspend function to a flow, I had to create an `emptyFlow()` and call `onStart`.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
nomisRev: @Intex32 I explain back-pressure a little bit here. https://youtu.be/n724AdSAkOI
Let me know if this clears it up for you, if not we can discuss it further ☺️

nomisRev (approved):
Looks good to me @Intex32! Great work 🙌

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
Intex32: @nomisRev I have more questions than before 😂. I especially liked the part where you talked about back pressure.
The pen has some classy dance moves tbh.

PR #388, opened 2023-09-05
update openai client to 3.4.1

raulraja:


openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAIClient.kt
```diff
@@ -58,9 +59,10 @@ class OpenAIModel(
       headers = mapOf("Authorization" to " Bearer $openAI.token")
     )
 
+  @OptIn(LegacyOpenAI::class)
```
nomisRev, line 62: Is this going to be removed? Or why does it require this annotation?

nomisRev (approved):

Montagon (approved):

openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAIClient.kt
raulraja: Open AI is fading out the completion API because the chat completion endpoint can encode the same capabilities. Eventually, it will disappear. 

https://platform.openai.com/docs/guides/gpt/completions-api

> The completions API endpoint received its final update in July 2023 and has a different interface than the new chat completions endpoint. Instead of the input being a list of messages, the input is a freeform text string called a prompt.

https://platform.openai.com/docs/guides/gpt/chat-completions-vs-completions
> The Chat completions format can be made similar to the completions format by constructing a request using a single user message. For example, one can translate from English to French with the following completions prompt:

> Likewise, the completions API can be used to simulate a chat between a user and an assistant by formatting the input [accordingly](https://platform.openai.com/playground/p/default-chat?model=text-davinci-003).

> The difference between these APIs derives mainly from the underlying GPT models that are available in each. The chat completions API is the interface to our most capable model (gpt-4), and our most cost effective model (gpt-3.5-turbo). For reference, gpt-3.5-turbo performs at a similar capability level to text-davinci-003 but at 10% the price per token! See pricing details [here](https://openai.com/pricing).

PR #411, opened 2023-09-08
Update Xef DB to incorporate users, projects and organizations

Montagon:
This PR includes a migration to the Xef DB to include users, projects and organizations.


javipacheco (dismissed):
🚀  LGTM! 

I have migrated the database locally and everything works as expected

server/src/main/kotlin/com/xebia/functional/xef/server/models/Database.kt
```diff
@@ -0,0 +1,16 @@
+package com.xebia.functional.xef.server.models
+
+import kotlinx.datetime.Instant
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class XefTokens(
```
javipacheco, line 8: Should we move this class to `XefTokensTable.kt`?

I know that it is a little bit different to others because of multiple primary keys, but I think that we should have this class with the table like `UsersTable` and `User`

javipacheco (approved):
🚀  LGTM! Really great job as always!

Just a minor comment, please let me know your opinion about that

server/src/main/kotlin/com/xebia/functional/xef/server/models/Database.kt
Montagon: Oook, seems good. Updated!

PR #426, opened 2023-09-14
Added endpoints for manage organizations

Montagon:


server/src/main/kotlin/com/xebia/functional/xef/server/services/OrganizationRepositoryService.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.jetbrains.exposed.sql.SizedCollection
+import org.jetbrains.exposed.sql.transactions.transaction
+import org.slf4j.Logger
+
+class OrganizationRepositoryService(
+    private val logger: Logger
+) {
+    fun createOrganization(
+        data: OrganizationRequest,
+        token: String?
```
javipacheco, line 17: For me, It's weird to send a nullable token here. I think that we can use this function in `XefRoutes`

```kotlin
private fun ApplicationCall.getToken(): String =
    principal<UserIdPrincipal>()?.name ?: throw IllegalArgumentException("No token found")
```

We should set it as a public function first, and use it in the endpoint. Then we can pass a `token: String` not nullable here

Montagon: Yep, it doesn't make sense. I will update it, thanks!

PR #430, opened 2023-09-15
[428] Better error handling in Xef Server

Montagon:


server/src/main/kotlin/com/xebia/functional/xef/server/exceptions/ExceptionsHandler.kt
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+}
+
+suspend fun ApplicationCall.manageException(cause: XefExceptions) {
+    when(cause) {
+        is XefExceptions.ValidationException -> this.respond(HttpStatusCode.BadRequest, cause.message)
+        is XefExceptions.AuthorizationException -> this.respond(HttpStatusCode.Unauthorized)
+        is XefExceptions.OrganizationsException -> this.respond(HttpStatusCode.BadRequest, cause.message)
+        is XefExceptions.UserException -> this.respond(HttpStatusCode.BadRequest, cause.message)
```
Montagon, line 30: Perhaps makes sense to have this more granular. To have different types of User exceptions, and return different statuses on each case. The same for the rest of the exceptions 🤔

Montagon:
This PR will be updated in order to work with Arrow.

server/src/main/kotlin/com/xebia/functional/xef/server/exceptions/ExceptionsHandler.kt
```diff
@@ -0,0 +1,32 @@
...
+
+fun Application.exceptionsHandler() {
+    install(StatusPages) {
+        exception<Throwable> { call, cause ->
+            if (cause is XefExceptions) {
+                call.manageException(cause)
+            } else {
+                call.respond(HttpStatusCode.InternalServerError, cause.message ?: "Unexpected error")
+            }
+            call.respondText(cause.localizedMessage, status = HttpStatusCode.InternalServerError)
```
Intex32, line 17: Why do we respond twice here? Once inside the if-else branch and once after that.

Montagon: If I'm not wrong, they belong to two different branches. 
```kotlin
install(StatusPages) {
        exception<Throwable> { call, cause ->
            if (cause is XefExceptions) {
                call.manageException(cause)
            } else {
                call.respond(HttpStatusCode.InternalServerError, cause.message ?: "Unexpected error")
            }
            call.respondText(cause.localizedMessage, status = HttpStatusCode.InternalServerError)
        }
        status(HttpStatusCode.NotFound) { call, status ->
            call.respondText(text = "404: Page Not Found", status = status)
        }
    }
```

The first is the `exception` handler. In which we are getting the exceptions and in the other case is the `status`, in which is changing the response of all the NotFound responses.

Montagon: I mistake 😓 Thanks!

javipacheco (approved):
🚀  LGTM!

I have tried the code locally and everything works as expected. We will address to include Arrow lib in a separate PR

Intex32:
> I have tried the code locally and everything works as expected. We will address to include Arrow lib in a separate PR

Are you going to use Arrow's Either and EitherNel for error handling? What are your plans?

Montagon:
> > I have tried the code locally and everything works as expected. We will address to include Arrow lib in a separate PR
> 
> Are you going to use Arrow's Either and EitherNel for error handling? What are your plans?

The plan is to apply the same approach as here: https://github.com/nomisRev/ktor-arrow-example/pull/35/files#diff-fd19be7abb78848f1cf6bc8da926adafd98399c75cf49662e9a8dcc29f7f3cb0R33

PR #438, opened 2023-09-18
Add PromptConfiguration ( temperature = 0.0 ) to ReActAgent

Zevleg:
Add temperature = 0.0 because the AI is creating tools that they don't exists


javipacheco (approved):
🚀

PR #439, opened 2023-09-19
Projects Endpoints

javipacheco:


server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/XefRoutes.kt
```diff
@@ -114,3 +114,9 @@ private fun ApplicationCall.getProvider(): Provider =
 
 fun ApplicationCall.getToken(): String =
     principal<UserIdPrincipal>()?.name ?: throw XefExceptions.AuthorizationException("No token found")
+
```
Montagon, line 117: :heart:

server/src/main/kotlin/com/xebia/functional/xef/server/services/RepositoryService.kt
```diff
@@ -20,3 +24,6 @@ object RepositoryService {
     }
 
 }
+
+fun String.getUser() =
```
Montagon, line 28: I like the idea, but in this case, I would say that makes more sense to wrap the String value in a Token class or something similar. I don't know if is good to add a function to a primitive type which depends on a BD to work.

Montagon (changes requested):
Should you update the postman collection in order to add those endpoints?

server/src/main/kotlin/com/xebia/functional/xef/server/services/RepositoryService.kt
javipacheco: Sure!

Montagon (approved):
🚀

PR #440, opened 2023-09-20
storing full prompt and response messages

Montagon:
This PR modifies the behaviour of storing messages in memory. We realized that we were skipping possible messages from the request. The behaviour before will be something like this:

RequestMessages = ["You are an expert on classical music", "Which is the best composition of Gustav Mahler?"]
AIResponseMessage = ["One of the best compositions is Mahler's Second Symphony"]

As a result of the previous behaviour, we will end with this in the memory:
["Which is the best composition of Gustav Mahler?", "One of the best compositions is Mahler's Second Symphony"]

We are missing all the request messages except the last one. This PR fixed that adding to the memories all the request messages.


javipacheco (approved):
🚀

PR #464, opened 2023-09-28
441 fine tuning oai

Intex32:
- xef-core: implemented an API for querying fine-tuned models
- xef-server: cloned OAI endpoints for fine tuning (to later intercept and collect metrics)

Example in `examples/kotlin/src/main/kotlin/com/xebia/functional/xef/conversation/finetuning/FineTunedModelChat.kt`

Current limitations:
- the abstract copy function from `LLM` has to return the most specific type of this instance, this cannot be ensured at compile time, in the `spawnModel` function the return value is casted where it could potentially fail
- it cannot be assured that the fined tuned model has the specified capability in `spawnModel`


Intex32:
Open for first discussions :)
Happy about feedback

Issue #466, opened 2023-09-29
Update Memory


javipacheco:
In order to fix some problems with the way to order the memory in the conversations, we need to change the way to store the memory

1. We have to remove the `timeStamp` and the `approxTokens`  in the `Memory`. The memory could be something like this:

```kotlin
data class Memory(val conversationId: ConversationId, val content: Message, val index: Int)
```

2. The index is an ordered number that the conversation is incrementally added to.

PR #467, opened 2023-09-29
Memory updated

javipacheco:
This PR closes issue #466

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MemoryManagement.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MemoryManagement.kt
@@ -23,7 +23,7 @@ internal fun Message.toMemory(cid: ConversationId, index: Int): Memory =
 internal fun List<Message>.toMemory(scope: Conversation): List<Memory> {
   val cid = scope.conversationId
   return if (cid != null) {
-    mapIndexed { index, it -> it.toMemory(cid, scope.store.incrementIndexAndGet()) }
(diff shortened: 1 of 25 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MemoryManagement.kt
```diff
@@ -4,85 +4,51 @@ import com.xebia.functional.xef.conversation.Conversation
...
-  )
+internal fun Message.toMemory(cid: ConversationId, index: Int): Memory =
+  Memory(conversationId = cid, content = this, index = index)
 
-internal fun List<Message>.toMemory(chat: LLM, scope: Conversation): List<Memory> {
+internal fun List<Message>.toMemory(scope: Conversation): List<Memory> {
   val cid = scope.conversationId
   return if (cid != null) {
-    mapIndexed { index, it -> it.toMemory(cid, chat, index) }
+    mapIndexed { index, it -> it.toMemory(cid, scope.store.incrementIndexAndGet()) }
```
Montagon, line 26: I think we are not using the index, right? Could we change it to `map`?

Montagon (approved):
Great job @javipacheco 👏

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MemoryManagement.kt
javipacheco: Good catch

raulraja (dismissed):
👍

Montagon (approved):

PR #468, opened 2023-09-29
Simplify ReAct agent and add streaming

raulraja:


Montagon (approved):
So much simpler now! 👏 🚀

PR #469, opened 2023-10-02
query fine tuned models

Intex32:
sub PR of #464 

Example in examples/kotlin/src/main/kotlin/com/xebia/functional/xef/conversation/finetuning/FineTunedModelChat.kt

Current limitations:
- the abstract copy function from LLM has to return the most specific type of this instance, this cannot be ensured at compile time, in the spawnModel function the return value is casted where it could potentially fail
- it cannot be assured that the fined tuned model has the specified capability in spawnModel


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/LLM.kt
```diff
@@ -12,6 +12,12 @@ sealed interface LLM : AutoCloseable {
   val name
     get() = modelType.name
 
+  /**
+   * Copies this instance and uses [modelType] for [LLM.modelType].
+   * Has to return the most specific type of this instance!
+   */
+  fun copy(modelType: ModelType): LLM
```
Intex32, line 19: I don't like this, but I didn't see a better way. Every topmost subclass of LLM must return an instance of it's type so that models based on the original model with the same capabilities is spawned.

tokenizer/src/commonMain/kotlin/com/xebia/functional/tokenizer/ModelType.kt
```diff
@@ -84,6 +84,9 @@ sealed class ModelType(
   object CODE_SEARCH_BABBAGE_CODE_001 : ModelType("code-search-babbage-code-001", R50K_BASE, 2046)
   object CODE_SEARCH_ADA_CODE_001 : ModelType("code-search-ada-code-001", R50K_BASE, 2046)
 
+  class FineTunedModel(name: String, val baseModel: ModelType)
```
Montagon, line 87: Could you format the code?

examples/kotlin/src/main/kotlin/com/xebia/functional/xef/conversation/finetuning/FineTunedModelChat.kt
```diff
@@ -0,0 +1,19 @@
+package com.xebia.functional.xef.conversation.finetuning
+
+import com.xebia.functional.xef.conversation.llm.openai.OpenAI
+import com.xebia.functional.xef.prompt.Prompt
+
+suspend fun main() {
+  val OAI = OpenAI()
+  val model =
+    OAI.spawnModel("ft:gpt-3.5-turbo-0613:xebia-functional:ron-v2:830qZBwQ", OAI.GPT_3_5_TURBO)
```
Montagon, line 9: I think we should not have this value here.

tokenizer/src/commonMain/kotlin/com/xebia/functional/tokenizer/ModelType.kt
Intex32: I ran spotless on it... Isn't this enough?

tokenizer/src/commonMain/kotlin/com/xebia/functional/tokenizer/ModelType.kt
```diff
@@ -84,6 +84,9 @@ sealed class ModelType(
...
+    : ModelType(name = name, encodingType = baseModel.encodingType, maxContextLength = baseModel.maxContextLength, tokensPerMessage = baseModel.tokensPerMessage, tokensPerName = baseModel.tokensPerName, tokenPadding = baseModel.tokenPadding)
+
```
Montagon, line 89:
```suggestion
  class FineTunedModel(name: String, val baseModel: ModelType) : ModelType(
    name = name,
    encodingType = baseModel.encodingType,
    maxContextLength = baseModel.maxContextLength,
    tokensPerMessage = baseModel.tokensPerMessage,
    tokensPerName = baseModel.tokensPerName,
    tokenPadding = baseModel.tokenPadding
  )

```

Montagon (approved):
🚀

openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAI.kt
```diff
@@ -138,6 +132,28 @@ class OpenAI(internal var token: String? = null, internal var host: String? = nu
       DALLE_2,
     )
 
+  suspend fun findModel(modelId: String): Any? { // TODO: impl of abstract provider function
```
javipacheco, line 135: This method should be private and return `ModelType?` not `Any?`

```suggestion
  private suspend fun findModel( modelId: String): ModelType?
```

Issue #405, opened 2023-09-08
abstract ModelType


Intex32:
Goal: generalize the ModelType class from OpenAI specific implementations to support multiple providers

I consider `ModelType` a formal description of a model and it's properties without any capabilities. This includes eg name, contextLength.

my suggestions:
- move file from module `tokenizer` to `core`
- remove sealed class constraint
- implementations of `ModelType` can be found directly in the provider specific classes that contain instances of available models; instance is passed as parameter to respective model constructor => move OpenAI instances and create new instances with correct parameters for gpt4all and GCP
- encodingType has to be moved out of `ModelType`; GCP appears to not have public information about the encoding they are using;  thus we have to make an API call to GCP to retrieve the token count of a given message; for OpenAI, the encoding is used to locally compute the token count

depends on #393

PR #477, opened 2023-10-04
405 abstract ModelType - inline ModelType

Intex32:
The idea of this PR is outlined in [#405 ](https://github.com/xebia-functional/xef/issues/405#issuecomment-1745098212).

In short, ModelType is currently too much designed for the OpenAI. By inlining all the properties and functionality of ModelType to the right interface in the LLM hierarchy, it can be made more generic and type safe.

What I did here:
- introduces a `ModelID` value class (kinda as a replacement for `ModelType` as this is the only feature all LLMs share
- inlines the model's name and context length - and the encoding for all OpenAI models into new interface `OpenAIModel`
- introduce `MaxIoContextLength` - OpenAI's models all have a shared limit for input and output tokens combined, some of Google's models have separate limits (one for input and output)
- the OpenAI specific implementation of `tokensFromMessages` had to be moved out of `LLM`; additionally `countTokens` and `truncateText` have moved to the same subclass of `LLM`
- changes the Tokenizer functionaliy to use Encoding directly rather than through the model's `modelType`

Besides the internal refactoring, tests and integrations had to be adapted.
- `TokenTextSplitter` now takes `EncodingType` rather than `ModelType` as parameter (as is still specific to OAI)
- some tests that were using the encoding type or context length had to be adapted

If this is approved, a subsequent PR building on these changes may be considered to abstract the encoding part for estimating the tokens. Part of this job is adapting internal APIs and integrations to the new more abstract `contextLength` (of type `MaxIoContextLength`).


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/BaseChat.kt
```diff
@@ -0,0 +1,25 @@
...
+
+  @Deprecated(
+    "will be removed from LLM in favor of abstracting former ModelType, use contextLength instead"
+  )
+  val maxContextLength
+    get() =
+      (contextLength as? MaxIoContextLength.Combined)?.total
+        ?: error(
+          "accessing maxContextLength requires model's context length to be of type MaxIoContextLength.Combined"
+        )
```
Intex32, line 18: As a side note, again - this is supposed to be an intermediary solution. Usages of this field that use an OAI model will still work as before. If this field is called on an instance of a (Google) model that doesn't use the combined context length an exception imminent.
I found this to be the best way to handle it, in favor to not change too much code in one PR.

Intex32:
Hey everyone 👋, I think this PR is ready for a first review. In the very infancy of this PR i discussed my ideas with @raulraja. This is what I came up with. Open for discussion :)
Since I have no GCP token anymore, somebody pls test if the GCP example is still working.

raulraja:
Hi Ron, we are getting rid of the GCP integrations in favor of having a single server compatible with the Open AI YAML spec based on the branch work in `use-generated-openai-client`. That branch actually gets rid of the entire LLM hierarchy and autogenerates the services based on the YAML spec.

PR #478, opened 2023-10-05
Integrates MLflow Gateway as an integration

fedefernandez:
Initial support for MLflow Gateway. Xef would need some refactoring to accommodate properly the way of working of this gateway. Open for discussion.

Next step will be to add the integration directly on the server


Intex32:
Hey Fede 👋! As we're gradually adding support for new providers such as GCP and now MLFlow, making xef-core generic gets more important. One of my overall all goals during my stay here was to push GCP to have the same support as OAI in XEF - which required - and still does - some refactoring.
What do you need in order for XEF to _properly work with Gateway_? Perhaps I can consider that in my work as I still have a similar open issue.

fedefernandez:
hey @Intex32, I started with the part that was more familiar to me (the integrations), but today, I talked with @raulraja, and we agreed it would be better to implement an approach for having the integration in the server. I have created a draft PR with a potential solution here: https://github.com/xebia-functional/xef/pull/479

PR #479, opened 2023-10-05
MLflow gateway and different OpenAI backends

fedefernandez:
This PR brings generics over the Xef server backend API. With the code shown in `server/src/main/kotlin/com/xebia/functional/xef/server/Server.kt`, you can easily switch from OpenAI to MLflow Gateway. 

We can generalize it further, read that from the configuration, and decide when to spin up the server.

The `OpenAI` now works with any "OpenAI gateway". See instructions below for testing with LiteLLM

## Pending items

Below are some items that are missing from this implementation

- [x] **Decide the backend API on server start**: It would be nice to decide the backend API when the server starts. This should come from the configuration. _It's optional for this PR to be completed._
- [x] **Implement `PathProvider`**: This item returns the path for the gateway given a model name and a `UserIdPrincipal`. Note: At this moment, the Xef server builds the `UserIdPrincipal` with the name equal to the OpenAI token. A potential implementation could be:
  1. On server startup, read all the MLflow gateways
  2. Create a map from model names to a list of paths
  3. Return the first path given a model name
- [ ] **Review architectural and code style**: I'm not very familiar with how the code in Xef is structured. I have organized it into different files and models. It would be nice to have a review on this angle.

#### Start LiteLLM Proxy Server

1. Install the library and start the server

```
pip install litellm
litellm --model huggingface/bigcode/starcoder
```

#### Start Xef server

1. Change `Server.kt` to start the `aiRoutes` with the new OpenAI url

```kotlin
aiRoutes(OpenAIApiProvider("http://0.0.0.0:8000", ktorClient))
```

2. Make some requests

**Note 1**: Xef needs the Bearer token here, but it is not used. It only checks that an authentication exists. **Send a mocked token, because you will be sending the token to `bigcode/starcoder`**.

**Note 2**: You can start LiteLLM without specifying a model and the one from the request will be used.

```bash
curl -v --location 'http://localhost:8081/chat/completions' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer XXX' \
--data '{
    "model": "gpt-3.5-turbo",
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant."
      },
      {
        "role": "user",
        "content": "Hello!"
      }
    ]
  }'
```

#### Start MLflow Gateway

1. Install the library

```
pip install 'mlflow[gateway]'
```

3. Export an OpenAI API Key

```
export OPENAI_API_KEY=your_api_key_here
```

5. Choose a directory and create a config file

```
mkdir -p ~/development/mlflow
cd ~/development/mlflow
vi config.yaml
```

Paste the following content

```yml
routes:
  - name: chat
    route_type: llm/v1/chat
    model:
      provider: openai
      name: gpt-3.5-turbo
      config:
        openai_api_key: $OPENAI_API_KEY

  - name: embeddings
    route_type: llm/v1/embeddings
    model:
      provider: openai
      name: text-embedding-ada-002
      config:
        openai_api_key: $OPENAI_API_KEY
```

6. Start the Gateway

```
mlflow gateway start --config-path config.yaml
```

#### Start Xef server

1. Change `Server.kt` to start the `aiRoutes` with the `MLflowApiProvider`

```kotlin
// Load the mlflow provider
val mlflowProvider = mlflowApiProvider("http://127.0.0.1:5000", ktorClient)

// Add it as a ai route
aiRoutes(mlflowProvider)
```

2. Make some requests

**Note**: The Bearer token here is needed by Xef but not used, it only check that an authentication exists.

```bash
curl -v --location 'http://localhost:8081/chat/completions' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer XXX' \
--data '{
    "model": "gpt-3.5-turbo",
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant."
      },
      {
        "role": "user",
        "content": "Hello!"
      }
    ]
  }'
```


server/src/main/kotlin/com/xebia/functional/xef/server/models/mlflow/JsonFields.kt
Intex32, line 1: Having these fields as public top level declarations is not a good practice IMO. Can we move them to the corresponding provider class perhaps?

server/src/main/kotlin/com/xebia/functional/xef/server/ai/providers/MLflowApiProvider.kt
```diff
@@ -0,0 +1,85 @@
...
+        val stringRequestBody = requestBody.toString(Charsets.UTF_8)
+        val requestData = Json.decodeFromString<JsonObject>(stringRequestBody)
+        val isStream = requestData["stream"]?.jsonPrimitive?.boolean ?: false
+        if (isStream) {
+            call.respond(HttpStatusCode.NotImplemented, "Stream not supported")
+        } else {
+            val model = requestData["model"]!!.jsonPrimitive.content
+            val principal = call.principal<UserIdPrincipal>() ?: throw XefExceptions.AuthorizationException("")
+            val path = pathProvider.chatPath(model, principal)
+            val newData = requestData.filterKeys { !filterOutFields.contains(it) }.mapKeys {mappedFields.getOrDefault(it.key, it.key) }
```
Intex32, line 37: I do not know exactly about the API specification and how it relates to the OAI spec.
Though, does it make more sense to specify _what fields are needed_ rather than which are not?

server/src/main/kotlin/com/xebia/functional/xef/server/ai/providers/mlflow/PathProvider.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import org.slf4j.Logger
+
+interface PathProvider {
+    fun chatPath(model: String, principal: UserIdPrincipal): String?
+    fun embeddingsPath(model: String, principal: UserIdPrincipal): String?
+}
+object DefaultPathProvider: PathProvider {
```
Intex32, line 21: What do we need this class for? It is not used

server/src/main/kotlin/com/xebia/functional/xef/server/ai/providers/MLflowApiProvider.kt
fedefernandez: I don't know. To give more context, mlflow gateway throws an error when you send some parameters that the API is not expecting (like the `model`). On the other hand, having a list of the needed fields will require us to change the list if new fields are added. With the current implementation, if a new field is added, the clients can start sending them immediately. Do you think it makes sense?

server/src/main/kotlin/com/xebia/functional/xef/server/ai/providers/mlflow/PathProvider.kt
fedefernandez: Previous code, I'll remove it 👍🏼

raulraja (approved):
As discussed over huddle, we'd like to go with this approach, thanks Fede!

fedefernandez:
@raulraja this is ready for reviewing

krrishdholakia:
Hey @fedefernandez -- you can run litellm without setting the model param 

```
litellm 
```

would spin up the proxy, and accept any of the models/providers we accept (huggingface tgi, anthropic, togetherai, replicate, etc.) by reading in the model name from the chat completions call

fedefernandez:
@krrishdholakia that's awesome, I'll update the description

krrishdholakia:
> It would be nice to have some streaming JSON mapper to transform JSON into different schemas. 

We also support streaming across all our providers 

https://docs.litellm.ai/docs/completion/stream#streaming-responses

---
@fedefernandez / @raulraja curious - what's missing from litellm to be used in prod?

fedefernandez:
Moving to draft, it will be divided into different pieces for a better review. The first one is #507

PR #481, opened 2023-10-06
Metrics with OpenTelemetry

javipacheco:
This PR is the first approach for adding metrics to Xef. 

The new `Metric` component is added to `Conversation` in the same way that we are adding `VectorStore`. The metrics contain 2 implementations:

- `LogsMetric`: it is the default implementation in `core`. This implementation adds logs in the console
- `OpenTelemetryMetric`: it is an integration with OpenTelemetry

Here is an image using `JaegerUI`

Screenshot 2023-10-06 at 13 27 31

You have comments in the `Animal.kt` example for testing OpenTelemetry locally


raulraja:
this is looking great!, thank you @javipacheco !

integrations/opentelemetry/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+    implementation(libs.opentelemetry.extension.kotlin)
+    implementation(libs.opentelemetry.exporter.otlp)
+
+    testImplementation(libs.junit.jupiter.api)
+    testImplementation(libs.kotest.property)
+    testImplementation(libs.kotest.framework)
+    testImplementation(libs.kotest.assertions)
+    testRuntimeOnly(libs.kotest.junit5)
+}
+
```
Montagon, line 41: I think you should add the spotless plugin here. Something like:

Inside `plugins`:
`alias(libs.plugins.spotless)`

And outside:

```gradle
spotless {
  kotlin {
    target("**/*.kt")
    ktfmt().googleStyle().configure {
      it.setRemoveUnusedImport(true)
    }
  }
}
```

integrations/opentelemetry/src/main/kotlin/com/xebia/functional/xef/opentelemetry/OpenTelemetryConfig.kt
```diff
@@ -0,0 +1,50 @@
...
+data class OpenTelemetryConfig(
+    val endpointConfig: String,
+    val defaultScopeName: String,
+    val serviceName: String
+) {
+
+    fun newInstance(): OpenTelemetry {
+        val jaegerOtlpExporter: OtlpGrpcSpanExporter = OtlpGrpcSpanExporter.builder()
+            .setEndpoint(endpointConfig)
+            .setTimeout(30, TimeUnit.SECONDS)
```
Montagon, line 22: Should be this parameter taken from config or environment?

Montagon (approved):
Just some comments. Great job, @javipacheco! 🚀

integrations/opentelemetry/src/main/kotlin/com/xebia/functional/xef/opentelemetry/OpenTelemetryConfig.kt
javipacheco: I think that it's not necessary for now

integrations/opentelemetry/build.gradle.kts
javipacheco: We should add spotless in a new PR

Montagon (approved):

PR #465, opened 2023-09-29
441 fine tuning intercepting endpoints

Intex32:
sub PR of #464 
cloned OAI endpoints for fine tuning (to later intercept and collect metrics)

PR #482, opened 2023-10-08
server: request forwarding

Intex32:
As it looks, #465 will not be resolved in near future. I would like to merge these changes into main that originated from #465.

This PR provides a file with with a function to forward an incoming request to a provider like OAI cloning parameters, headers, path, http method and body.
The previous implementation was too specific to it's usages, involved some unnecessary boilerplate and was made more generic here. Existing endpoints were adapted.


server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/RequestForwardingUtil.kt
```diff
@@ -0,0 +1,97 @@
...
+        }
+    } else {
+        respond(providerResponse.status, providerResponse.readBytes())
+    }
+}
+
+private fun ResponseHeaders.copyFrom(headers: Headers) = headers
+    .entries()
+    .filter { (key, _) -> !HttpHeaders.isUnsafe(key) } // setting unsafe headers results in exception
+    .filter { (key, _) -> !key.equals("Content-Length", ignoreCase = true) }
```
fedefernandez, line 90: I had the same problem in my PR. Do you know why it's needed here?

fedefernandez (dismissed):
Code looks good to me, just a minor comment

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/RequestForwardingUtil.kt
Intex32: I think you're right with your PR's comment. Ktor adds it when automatically when you set the response body. BUT, it used to work before without filtering out the header. Idk what changed. Yesterday when I tried it again, I got duplicate headers error by postman.

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/RequestForwardingUtil.kt
```diff
@@ -0,0 +1,97 @@
...
+    response.headers.copyFrom(providerResponse.headers)
+
+    if(stream) {
+        respondOutputStream {
+            providerResponse
+                .bodyAsChannel()
+                .copyTo(this@respondOutputStream)
+        }
+    } else {
+        respond(providerResponse.status, providerResponse.readBytes())
```
fedefernandez, line 83: @Intex32, is this reading the body twice (https://github.com/xebia-functional/xef/pull/482/files#diff-5767cc2ab5cd1d7765af84cd0859fec6baf453ecef7e81b518f58e0890a4b692R43)? The same for the streaming

fedefernandez (commented):
I have tried to incorporate these changes into my branch because I like the idea of forwarding. Still, we're moving towards having a difference between the Xef request/response model and the "backend" provider (OpenAI, MLflow, etc.) request/response model. Is it possible to incorporate some of these changes into #479? Happy to jump into a call.

server/src/main/kotlin/com/xebia/functional/xef/server/http/routes/RequestForwardingUtil.kt
Intex32: Yes, first your reading and parsing it to a JsonObject (for monitoring or so) and in the `forward` function the input is received a second time as raw bytes and passed forward. I found this to be the cleanest solution, as both is independent.

Intex32:
> Still, we're moving towards having a difference between the Xef request/response model and the "backend" provider (OpenAI, MLflow, etc.) 

At the time being, the only thing you do in the OpenAIProvider is forwarding, no? You do not transform the input, as far as I see. When merging you can adjust it to you needs, based off my work here.

fedefernandez:
> the only thing you do in the OpenAIProvider is forwarding, no?
At this moment, if I'm not wrong we want to have a compatible API, but nothing is stopping us from receiving more fields than the ones needed by OpenAI to do extra things.

> What tranformations are we talking about?
These ones: https://github.com/xebia-functional/xef/pull/479/files#diff-50cfa95ea6a3bb663c42091e510ab3c6373106f38c357a551f3e73434112d252R12-R40

Intex32:
I added some optional lambdas for the `forward` function that can transform - among other things - the req and res body. https://github.com/xebia-functional/xef/pull/482/commits/f4cec94c99fdc3fa9bf8f298b653a6b275cb8d30
Have a look if this helps you, if you think your use case does not fit my PR, I understand if you do it your own custom way.
image

fedefernandez:
@Intex32 I incorporated your changes in my branch (https://github.com/xebia-functional/xef/compare/feature/mlflow-routes-request-forwarding?expand=1), but then I started to test the code and found that the server responds with binary data, see the screenshot taken with a request executed using this branch. Are you seeing this behavior too?

Screenshot 2023-10-10 at 19 46 38

Intex32:
Dealing with the content type is messed up. Can you check again with my latest commit?
Sorry I didn't have much time to test the code.

fedefernandez:
@Intex32 I'm afraid I'm struggling with the same error. I see the benefits of this new approach, but it could conflict with the one we're following in the other PR. Do you mind if we try the other way around? by prioritizing #479 and then improving the connection there?

raulraja:
Moving this to DRAFT until the changes for the server in #479 are stable

PR #490, opened 2023-10-11
Fix - GenericIOException: Failed to parse HTTP response: unexpected EOF

tomascayuelas:
### What?
Testing our library in a simple use case we have detected that with a newly generated personal OPENAI_TOKEN. When we use it we get this exception 
`(com.aallam.openai.api.exception.GenericIOException: Failed to parse HTTP response: unexpected EOF)`

### Why?
Because the Timeout establishes a default time of 30 seconds for the socket, so if this request spend more time that this value (socket = 30.seconds) the execution thrown an exception.

### How?
We have added a configurable Timeout with some default values in our OpenAI class. This solves the timeout problem given by the "com.aallam" OpenAI library client, issue already closed https://github.com/aallam/openai-kotlin/issues/234

### Other minor changes:
We take the opportunity to do this little refactor:
- set the findModel function to private
- put below the private function findModel (We think that is more readable in this way)

### DIFF
--- a/openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAI.kt
+++ b/openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAI.kt
@@ -35,7 +35,7 @@ class OpenAI(
   internal var timeout: Timeout = Timeout.default()
 ) : AutoCloseable, AutoClose by autoClose() {
 
-  class Timeout private constructor(
+  class Timeout(
     val request: Duration,
     val connect: Duration,
     val socket: Duration,
--- a/openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAI.kt
+++ b/openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAI.kt
@@ -41,7 +41,7 @@ class OpenAI(
     val socket: Duration,
   ) {
     companion object {
-      private val REQUEST_TIMEOUT = 30.seconds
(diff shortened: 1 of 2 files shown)


Montagon (dismissed):
🚀

openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAI.kt
```diff
@@ -22,12 +22,32 @@ import kotlin.jvm.JvmField
...
 
-class OpenAI(internal var token: String? = null, internal var host: String? = null) :
-  AutoCloseable, AutoClose by autoClose() {
+class OpenAI(
+  internal var token: String? = null,
+  internal var host: String? = null,
+  internal var timeout: Timeout = Timeout.default()
+) : AutoCloseable, AutoClose by autoClose() {
+
+  class Timeout private constructor(
```
raulraja, line 38: Can we relax the `private constructor` to just public? otherwise, a user would never be able to construct a time out without using the `Timeout.default()`. In practice, the time-out policy depends on the use case. The current default timeout is a good policy for a server, but a user-facing client or a different use case may not want to wait minutes for connection or socket timeouts.

raulraja (dismissed):
Thank you @tomascayuelas! Looks great!

openai/src/commonMain/kotlin/com/xebia/functional/xef/conversation/llm/openai/OpenAI.kt
tomascayuelas: Totally agree, depends of the use case, so is better remove the private constructor. Thanks for the suggestion! :metal:

PR #491, opened 2023-10-12
Fix format violation reported by "Spotless" merging into main branch

tomascayuelas:
Fix a violation reported by spotless.

_My mistake I forgot to past this checks in local ;(_


raulraja:
The biggest issue is that `spotless` should have complained before we merged it into the main. The CI pipeline continues to be broken in that it lets in `main` code that does not get checked through spotless.

PR #503, opened 2023-10-20
Update parameters type in CFunction to JsonObject

Montagon:
The current definition of `CFunction` contemplates the `parameters` as a string. But, when trying to generate data for fine-tuning, the types are not correct. Now the parameters property is a JsonObject as it is in OpenAI.


core/src/jvmMain/kotlin/com/xebia/functional/xef/conversation/PlatformConversation.jvm.kt
```diff
@@ -58,12 +63,15 @@ actual constructor(
...
-  fun chatFunction(target: Class<*>): CFunction =
-    CFunction(
+  fun chatFunction(target: Class<*>): CFunction {
+    val targetString = JacksonSerialization.schemaGenerator.generateSchema(target).toString()
+    return CFunction(
       name = target.simpleName,
       description = "Generated function for ${target.simpleName}",
-      parameters = JacksonSerialization.schemaGenerator.generateSchema(target).toString()
+      parameters = Json.parseToJsonElement(targetString).jsonObject
     )
```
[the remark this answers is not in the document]
raulraja: I guess this is ok because in this case we generate the schema specifically for JVM cases only when you provide a Class. This only affects the Java and Scala serializer and bypasses the kotlinx serialization. As an alternative, we would have to find a way that given a `Class` we can get back a JSON element from kotlin serialization and therefore avoid Jackson. 
The alternative is to implement a serialization stack for `Class` or find a library that already does this if we don't want to rely on Jackson.
```kotlin
@OptIn(InternalSerializationApi::class)
inline fun Class<*>.jsonSchema(): String =
  encodeJsonSchema(buildSerialDescriptor(
    serialName = this.simpleName,
    kind = if (isEnum) SerialKind.ENUM else SerialKind.CONTEXTUAL,
    typeParameters = emptyArray()
  ) {
    declaredFields.forEach { 
      element(it.name, ...)
    }
  })
```
I think your solution while it has to deserialize to serialize again to JsonObject is fine for now.

raulraja (approved):
Thank you @Montagon !

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/functions/JsonSchema.kt
```diff
@@ -111,7 +111,7 @@ fun encodeJsonSchema(descriptor: SerialDescriptor): String =
   Json.encodeToString(JsonObject.serializer(), buildJsonSchema(descriptor))
 
 /** Creates a Json Schema using the provided [descriptor] */
-private fun buildJsonSchema(descriptor: SerialDescriptor): JsonObject {
+fun buildJsonSchema(descriptor: SerialDescriptor): JsonObject {
```
fedefernandez, line 114: Should we remove `encodeJsonSchema` since it's not used now?

core/src/jvmMain/kotlin/com/xebia/functional/xef/conversation/PlatformConversation.jvm.kt
```diff
@@ -58,12 +60,14 @@ actual constructor(
...
 
-  fun chatFunction(target: Class<*>): CFunction =
-    CFunction(
+  fun chatFunction(target: Class<*>): CFunction {
+    val targetString = JacksonSerialization.schemaGenerator.generateSchema(target).toString()
+    return CFunction(
       name = target.simpleName,
       description = "Generated function for ${target.simpleName}",
-      parameters = JacksonSerialization.schemaGenerator.generateSchema(target).toString()
+      parameters = Json.parseToJsonElement(targetString).jsonObject
```
fedefernandez, line 68: The current implementation (generating schema with Jackson, print to JSON string, and parsing with Kotlinx serialization makes me uncomfortable 😅), but at this point I think it's the right way.

fedefernandez (approved):

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/models/functions/JsonSchema.kt
Montagon: Removed, thanks

core/src/jvmMain/kotlin/com/xebia/functional/xef/conversation/PlatformConversation.jvm.kt
Montagon: Yes, I feel the same, but as Raul mentioned to me, I think this will change in a future PR.

fedefernandez (approved):

PR #507, opened 2023-10-25
MLflow Gateway models

fedefernandez:
## MLflow Gateway integration

This PR consists of three parts. The client and its models are defined in a new integration module. Then, the example uses the new integration.

### MLflow Gateway Models

`integrations/mlflow/src/commonMain/kotlin/com/xebia/functional/xef/mlflow/models.kt`

Includes the models for interacting with MLflow Gateway:

* Routes definitions: Include routes information like URI, model names, model type, etc...
* Chat, completion, and embeddings: Requests and responses for interacting with the exposed models

### MLflow Gateway Client

`integrations/mlflow/src/commonMain/kotlin/com/xebia/functional/xef/mlflow/MlflowClient.kt`

Client with some methods for interacting with the gateway and its models.

* searchRoutes
* getRoute
* prompt
* chat
* embeddings

### MLflow Gateway Example

`examples/kotlin/src/main/kotlin/com/xebia/functional/xef/conversation/mlflow/Example.kt`

Example inspired by the example shown in the following talk by Corey Zumar and Ben Wilson

* https://youtu.be/WpudXKAZQNI?si=RaSSfwe9cEWQCJXg&t=2164


raulraja (approved):
thank you @fedefernandez !

PR #508, opened 2023-10-26
Added empty metrics by default

Montagon:
This PR adds a Metric.EMPTY that will not print anything to the console. And it's the default behaviour.


examples/kotlin/src/main/kotlin/com/xebia/functional/xef/conversation/conversations/Animal.kt
```diff
@@ -29,7 +30,7 @@ suspend fun main() {
     messagePolicy { addMessagesFromConversation = MessagesFromHistory.NONE }
   }
 
-  OpenAI.conversation {
+  OpenAI.conversation(metric = LogsMetric()) {
```
franciscodr, line 33: Do we want to use `LogMetric()` here? Or should we change it to `Metric.EMPTY`?

Montagon: In this case, as it's an example, I think we can use this to see the logs.

franciscodr (approved):

PR #523, opened 2023-11-07
OpenTelemetry integration updated

javipacheco:
This PR improves the metrics integration in Xef and updates the OpenTelemetry implementation for using nested traces

Jagger Screenshot

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/Chat.kt
@@ -46,7 +46,7 @@ interface Chat : LLM {
 
   @AiDsl
   suspend fun promptMessages(prompt: Prompt, scope: Conversation): List<String> =
-    scope.metric.promptSpan(scope, prompt) {
+    scope.metric.promptSpan(prompt) {
       val promptMemories = prompt.messages.toMemory(scope)
       val adaptedPrompt =
         PromptCalculator.adaptPromptToConversationAndModel(prompt, scope, this@Chat)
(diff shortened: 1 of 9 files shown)


fedefernandez (dismissed):
Left some minor comments, looks great

fedefernandez (approved):

organize:
Sorry for grave-digging, but I saw this mentioned in the 0.0.4 changelog 🎉 

Tried it out and I have some questions involving the nested traces. Is there some reason for using `makeCurrent()` instead of  passing the span around as a coroutine context? What I mean is something like this:

```
val currentContext = coroutineContext.getOpenTelemetryContext()
val span = tracer.spanBuilder(name).setParent(currentContext).startSpan()
...
withContext(span.asContextElement()) {
 block()
}
...
```
Similarly, `event` and `setAttribute` could just retrieve the Otel `Context` from the context intrinsic if they're made suspending. I'm asking because the documentation for `makeCurrent()` explicitly states this:

> The default implementation of this method will store the ImplicitContextKeyed in a ThreadLocal. Kotlin coroutine users SHOULD NOT use this method as the ThreadLocal will not be properly synced across coroutine suspension and resumption. Instead, use withContext(value.asContextElement()) provided by the opentelemetry-extension-kotlin library.

I think the span state can be omitted entirely, and `block` becomes `block: suspend () -> A`. Thoughts?

Also, what happens if invoking `block` throws? Does it leak the span here? From my experiments elsewhere, the span status needs to be set explicitly with something like:
```
    try {
      block()
    } catch (err: ...) {
      span.setStatus(StatusCode.ERROR)
      span.recordException(err)
      throw err
    } finally {
      span.end()
    }
 ```
 
 Thanks for the awesome library ❤️

PR #526, opened 2023-11-09
Improve SQL module

### LINKED ISSUE #500: Improve the SQL module
We are interested in adding more capabilities to the existing module. Some requirements:

## Tabular data model

- Define a data model to represent tabular query results very generically.
  - No matter the nature of the SQL query, the result should be expressed in terms of this new model. 
  - It should contain a list of columns with name and type
  - It should contain a list of rows where each row matches the columns' specifications.

## DB Interaction

- Expose an operation that receives a query in string format, performs it, and returns the result in the tabular format mentioned above.
- Uses Exposed to interact with the database.

## LLM Interaction

- Expose an operation that receives these arguments:
  - DB connection config
  - Existing conversation
  - List tables (list of strings)
  - Description of the nature of some fields to provide more context. Like the possible values if they are enumerations, etc (a string)
  - The user input
- And performs:
  - Connects with the DB to retrieve the schema of the mentioned tables
  - Inject the relevant schema in a new Prompt
  - Inject the description of fields provided as arguments
  - Run the call to the LLM to infer the output.
- And finally returns an object that includes:
  - The SQL query that satisfies the input
  - A friendly response for the user (with wildcards like XXX to replace the final values in the response)
  - Optionally, a detailed query that provides more details about the involved items in the Main Query.

gerson24:
This PR closes #500 and proposes significant enhancements to the SQL module, including the ability to execute SQL queries and generate responses in a tabular format. 
The changes made are as follows:
 - A tabular data model consisting of columns and rows has been defined, allowing the generic representation of SQL query results.
- The module now interacts with the database using Exposed to obtain the schema data from the table names provided by the user. This eliminates the need to include the table schema information in the context.
- A function has been implemented that from a prompt, the table names and a context, returns a response with the queries and the resulting data. It also supports prompts about information about tables, about what questions you can ask or anything related to SQL.
- Code example have been included.

Documentation: https://github.com/xebia-functional/xef-website/pull/24


raulraja (approved):
Looks great @gerson24 @tomascayuelas , thank you!

rafaparadela (approved):
Great work. Thanks

PR #528, opened 2023-11-09
[Fix] - Update deprecated detekt source setter

tomascayuelas:
This PR remove all this warning messages:

```'setter for source: ConfigurableFileCollection' is deprecated. Setter will be removed in a future release. Use `from` or `setFrom` instead.```


gerson24 (approved):

Montagon (approved):
I had the same feeling with this warning, thanks for fixing it! 👏

Issue #527, opened 2023-11-09
Remove support for Java and Scala


raulraja:
While attempting to support multiple languages, we found that maintaining them has become a chore, and they have never been on par with Kotlin regarding the features we offered there.
This is mainly related to all the infra and additional functions and tricks needed to support `suspend` functions in targets like Java and Scala.
We will remove support for all languages but Kotlin until we have a stable API. At that point, we will support them through Codegen or other means.

Expectations

- Remove all files and examples related to Java and Scala.
- Update README, docs, and website.

PR #530, opened 2023-11-09
Remove support for Java and Scala

tomascayuelas:
Issue: https://github.com/xebia-functional/xef/issues/527

This PR remove all the support for Java and Scala, tell me if there are something that I forgot.


raulraja (approved):
This looks good and I think we should get it merged ASAP so we can proceed with the Open AI client refactor work without all these modules in the way.

PR #531, opened 2023-11-10
OpenAI api client gen

fedefernandez:
This PR:
* Adds a new module for generating the client through a grade task
* Add a fix for making the code fully KMP compilable (including JS)
* Fix the `ChatCompletionRequestMessage` model


raulraja (approved):
Looks great Fede!

PR #533, opened 2023-11-10
Adds a custom task for downloading the API

fedefernandez:
This PR adds a new custom task for downloading the API specification.

This also stores the current commit associated with the API.


raulraja (approved):
🚀

PR #535, opened 2023-11-13
Replace host, port, and database properties by url in VectorStore configuration

franciscodr:
The vector store configuration class requires the host, port, and database name to build the URL. This approach is more flexible for cases where the URI is defined differently, like databases stored in Google Cloud SQL.


javipacheco (approved):
🚀

PR #539, opened 2023-11-14
[DOC] - Fix some documentations

tomascayuelas:
Due to the remove java and scala task, we need to adapt the docs. So this PR fix that.

The intro.md is very old, so I decided to put the quickstart section from the website fixed instead of rewrite all this intro.md


docs/intro.md
```diff
@@ -1,52 +1,51 @@
...
+By default, the `OpenAI.conversation` block connects to [OpenAI](https://platform.openai.com/).
+To use their services you should provide the corresponding API key in the `OPENAI_TOKEN`
+environment variable, and have enough credits.
 
-In the next examples we'll write functions that rely on `ai`'s DSL functionality, 
-but without actually extracting the values yet using `getOrThrow` or `getOrElse`.
-We'll eventually call this functions from an `ai` block as we've shown above, and
-this allows us to build larger pipelines, and only extract the final result at the end.
+```shell
+env OPENAI_TOKEN=<your-token> <gradle-command>
```
Montagon, line 26: should we include here the option in which we pass the token directly?

docs/intro.md
```diff
@@ -108,33 +160,36 @@ often want to supplement the LLM with more data:
...
-These additional pieces of information are called the _contextScope_ in xef.ai, and are attached
+These additional pieces of information are called the _context_ in xef.ai, and are attached
 to every question to the LLM. Although you can add arbitrary strings to the context at any
 point, the most common mode of usage is using an _agent_ to consult an external service,
 and make its response part of the context. One such agent is `search`, which uses a web
 search service to enrich that context.
 
 ```kotlin
-import com.xebia.functional.xef.auto.*
+import com.xebia.functional.xef.reasoning.serpapi.Search
```
Montagon, line 170: for this, we need to set up an environment variable called: `SERP_API_KEY`. Perhaps we should add some information about it

Montagon (dismissed):
Just minor comments 🙌

docs/intro.md
tomascayuelas: @Montagon I don't know if is necessary in this point, but we can add it if you see that is interesting

Montagon (approved):

docs/intro.md
tomascayuelas: Add it! Thanks @Montagon 👏

docs/intro.md
Montagon: It was just an idea, it's fine for now. thanks!

PR #543, opened 2023-11-15
[FIX] - Gradle problem in buildAndTestSinglep and buildAndTestMultip

tomascayuelas:
Fix the problem with Gradle executing more than two exec commands.

A [Gradle Exec task](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html) will only run the last ‘commandLine’ defined inside its block. Putting multiple entries inside its block will not run multiple commands.

Before we only execute the xef-tokenizer module due to this behaviour. Now we execute all the modules.

I want to continue investigating if we can reduce times.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -57,10 +57,6 @@ fun includeOrNotModulesToCommand(modules: List<String>, platform: String, includ
     }.toTypedArray()
 }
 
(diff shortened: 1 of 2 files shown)


Montagon (approved):
Great find, @tomascayuelas!! 🚀

fedefernandez (approved):

PR #544, opened 2023-11-16
[CI] - Spotless GitHub actions

tomascayuelas:
- Spotless Apply
- Remove spotless command execution in: buildAndTestMultip and buildAndTestSinglep
- Automatic commit of formatted files using GitHub Actions

I think that can be interesting to have as the same way that we already do in arrow project.


Montagon (approved):
🚀

PR #547, opened 2023-11-17
Update ChatWithFunctions.kt

fedefernandez:
The current loop never gets `if (currentAttempts == maxDeserializationAttempts)` satisfied. This shifts the index to the right, so the condition is satisfied when the maximum is reached.


raulraja (commented):
Maybe we should use here rangeTo https://kotlinlang.org/docs/ranges.html

fedefernandez:
Makes sense, I'll take a look and adapt the code 👍

fedefernandez:
@raulraja the code has been adapted. Thanks

raulraja (approved):

Montagon (approved):

PR #548, opened 2023-11-20
Evaluator

javipacheco:
This PR adds the first approach for an evaluator. The evaluator is a tool that can be used to evaluate the performance of a model on a dataset. In this case, we are using [deepeval](https://docs.confident-ai.com/) as evaluation framework for LLMs.

Follow the README to install dependencies and use the Gradlew Task to generate a website with the results like this:

Screenshot 2023-11-22 at 10 22 15


raulraja (dismissed):
Thank you @javipacheco looks great!

evaluator-example/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Usage
+
+To try this example, you can run the following command:
+
+```bash
+./gradlew evaluator
+```
+
+After running the command, you will see the following output on
```
Montagon, line 36:
```suggestion
After running the command, you will have the results saved in a web, that you can see opening the file:
```

evaluator/src/main/kotlin/com/xebia/funcional/xef/evaluator/TestBuilder.kt
```diff
@@ -0,0 +1,63 @@
+package com.xebia.funcional.xef.evaluator
+
+import com.xebia.funcional.xef.evaluator.models.OutputDescription
+import kotlin.jvm.JvmSynthetic
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+
+class TestBuilder(val description: String, val metric: String) {
```
Montagon, line 10: Just a comment. Should this class be changed to `SuiteBuilder` or something similar? It sounds strange to me that the name `TestBuilder` is not a proper test itself.

Montagon (approved):
Just minor comment and a suggestion. Great job, @javipacheco!

Montagon (approved):

PR #561, opened 2023-11-24
Improves polymorphic serialization + tests

fedefernandez:
This PR:
* Moves models to individual files
* Completes deserializer for the polymorphic types
* Add tests for polymorphic serializations
* Fixes some default args (default args could be omitted from the Json representation by the serializer, causing issues)


raulraja (approved):
🚀

Montagon (approved):
🚀

PR #568, opened 2023-11-29
[xef-evaluator] - Improvements

tomascayuelas:
Improvements:
- Content validation
- Richer domain
- Change DSL
- Adding tests

Don't hesitate to comment anything.


evaluator-example/src/main/kotlin/com/xebia/functional/xef/evaluator/examples/TestExample.kt
```diff
@@ -3,44 +3,32 @@ package com.xebia.functional.xef.evaluator.examples
...
-          +OutputResponse("The movie is Jurassic Park")
+        itemSpec("Recipe for a chocolate cake") {
+          contextDescription { "Contains instructions for making a cake" }
+          outputResponse { OpenAI.conversation { promptMessage(input) } }
+          outputResponse { "The movie is Jurassic Park" }
         }
       }
-
-    file.writeText(spec.toJSON())
+      .toJsonFile(output, "data.json")
```
javipacheco, line 31: :clap:

javipacheco (approved):
🚀  Great job!!

I have tested the evaluator locally and everything works as expected.

fedefernandez (commented):
This looks great @tomascayuelas. Just one comment, how hard would be to use the `kotlin.test` framework instead `kotest`? In Arrow they followed the approach of using kotest just for the assertions due to compatibility reasons with the platform.

fedefernandez (dismissed):

tomascayuelas:
> This looks great @tomascayuelas. Just one comment, how hard would be to use the `kotlin.test` framework instead `kotest`? In Arrow they followed the approach of using kotest just for the assertions due to compatibility reasons with the platform.

Thanks @fedefernandez I'll check it and see if it's simple.

fedefernandez:
@tomascayuelas I made a migration in this commit, in case it helps
https://github.com/xebia-functional/xef/pull/561/commits/2c5a8528d71345f8db8514377bc0575befec98f0
Arrow's closed related PRs are also a good source of info

evaluator/src/main/kotlin/com/xebia/functional/xef/evaluator/EitherOps.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import arrow.core.Either
+import arrow.core.NonEmptyList
+import com.xebia.functional.xef.evaluator.models.SuiteSpec
+import com.xebia.functional.xef.evaluator.models.errors.ValidationError
+import java.io.File
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+
+object EitherOps {
```
fedefernandez, line 11: Question, in Kotlin is this a good practice (nested the functions in an object)? I have seen in the project the extension functions defined at top level

evaluator/build.gradle.kts
```diff
@@ -18,13 +18,22 @@ java {
...
 
 detekt {
-    toolVersion = "1.23.1"
-    source.setFrom(files("src/main/kotlin"))
-    config.setFrom("../config/detekt/detekt.yml")
-    autoCorrect = true
+  toolVersion = "1.23.1"
+  source.setFrom(files("src/main/kotlin"))
+  config.setFrom("../config/detekt/detekt.yml")
+  autoCorrect = true
```
fedefernandez, line 36: Bad intent?

evaluator/src/main/kotlin/com/xebia/functional/xef/evaluator/models/errors/SuiteSpecError.kt
```diff
@@ -0,0 +1,58 @@
+package com.xebia.functional.xef.evaluator.models.errors
+
+sealed interface ValidationError {
```
fedefernandez, line 3: Should this file be called `ValidationError` instead?

evaluator/src/test/kotlin/com/xebia/functional/xef/evaluator/EitherOpsSpec.kt
```diff
@@ -0,0 +1,91 @@
...
+import kotlin.io.path.absolutePathString
+import kotlin.io.path.createTempDirectory
+import kotlin.io.path.listDirectoryEntries
+import kotlin.test.Test
+import kotlinx.coroutines.test.runTest
+
+class EitherOpsSpec {
+  private val validSuiteSpec = runBlocking {
+    SuiteSpec(simpleString.next()) {
+      outputDescription { simpleString.next() }
```
fedefernandez, line 24: Each time this test is executed, it runs once with random data, right? In that case, I think it would be better to use fixed values because, right now, can lead to flaky tests. Not a big deal though, just a preference.

evaluator/src/test/kotlin/com/xebia/functional/xef/evaluator/ItemSpecBuilderSpec.kt
```diff
@@ -0,0 +1,165 @@
...
+        contextDescription { myContextDescription }
+        outputResponse { firstOutputResponse }
+        outputResponse { secondOutputResponse }
+      }
+
+    invalidItemSpec shouldBeLeft listOf<ValidationError>(EmptyOutputResponse, EmptyOutputResponse)
+  }
+
+  @Test
+  fun shouldInvalidWhenAnyCombinationOfInputAreEmpty() = runTest {
```
fedefernandez, line 149: All the invalid tests could be replaced by a property test that generates a tuple of the invalid `ItemSpec` + the error list.

evaluator/src/test/kotlin/com/xebia/functional/xef/evaluator/SuiteBuilderSpec.kt
```diff
@@ -0,0 +1,205 @@
...
+          outputResponse { simpleString.next() }
+        }
+
+        itemSpec(simpleString.next()) {
+          contextDescription { simpleString.next() }
+          outputResponse { simpleString.next() }
+          outputResponse { emptyOutputResponse }
+        }
+      }
+      .shouldBeLeft() shouldBe listOf<ValidationError>(EmptyItemSpecOutputResponse(1))
```
fedefernandez, line 203: Same as above. I prefer the way `ItemSpecBuilderSpec` defines the data, but here we can also group some tests in a single one.

evaluator/src/test/kotlin/com/xebia/functional/xef/evaluator/utils/Generators.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import io.kotest.property.Arb
+import io.kotest.property.arbitrary.Codepoint
+import io.kotest.property.arbitrary.az
+import io.kotest.property.arbitrary.string
+import io.kotest.property.arbitrary.stringPattern
+
+object Generators {
+  val simpleString = Arb.string(1, 60, Codepoint.az())
+  val emptyString = Arb.stringPattern("\\s+")
```
fedefernandez, line 11: I think this could be probably more efficient to generate a random-size list of one of the blank spaces

evaluator/src/test/kotlin/com/xebia/functional/xef/evaluator/TestModelsSpec.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.xebia.functional.xef.evaluator.models.errors.EmptyOutputResponse
+import com.xebia.functional.xef.evaluator.utils.Generators.emptyString
+import com.xebia.functional.xef.evaluator.utils.Generators.simpleString
+import io.kotest.assertions.arrow.core.shouldBeLeft
+import io.kotest.assertions.arrow.core.shouldBeRight
+import io.kotest.property.arbitrary.next
+import kotlin.test.Test
+import kotlinx.coroutines.test.runTest
+
+class TestModelsSpec {
```
fedefernandez, line 17: These can me coded as a property-based tests

fedefernandez (commented):
Left some minor comments. I think it would be worthy to address or discuss

tomascayuelas:
Thank you very much @fedefernandez for your comments/review.
As we have already spoken by slack, I take care of them, I think they are very good, and can be better code. Thank you!

evaluator/src/main/kotlin/com/xebia/functional/xef/evaluator/models/errors/SuiteSpecError.kt
tomascayuelas: I moved the ValidationError to a new file.

nomisRev (changes requested):
We're here using Kotest inside Kotlin Test 🤕 Never seen that before 😅 Why are we using both?

We can also define a simple DSL by ourselves that covers the requirements of this PR. It's probably not much more than what we've already defined on top of Kotest.

Additionally, I agree with @fedefernandez that we should either properly do prop testing (with random/configurable SEED), or use fixed data.

A couple small questions, thoughts, and concerns on some the domain classes defined in this PR.

fedefernandez:
> @fedefernandez @tomascayuelas is this still relevant?

@raulraja I feel it's. I can take a look and prepare for reviewing

> We're here using Kotest inside Kotlin Test 🤕 Never seen that before 😅 Why are we using both?

We faced some binary issues and found the issues you created in Arrow. We're using the kotlin test framework but kotest for the assertions.

https://github.com/arrow-kt/arrow/issues/3183

nomisRev:
I see I am blocking this PR @fedefernandez.

To give some context on Arrow, that was to avoid rewriting more code and to political continue supporting Kotest. We replaced the Kotest runner only because we wanted to support WASM, and be able to build EAP releases (as part of the community builds).

However, I don't want to block this PR on that matter. I do feel we can simplify some things here, Junit also support "property based testing" in the form of table based testing. Not sure if Qwik works for KMP, but I've used it for kotlin-test on JVM.

That leaves the same concerns as you originally raised @fedefernandez concerning testing practices. I am removing my request for changes however to unblock this PR from my side.

fedefernandez:
Thanks @nomisRev. I have already started resolving some comments; I'll get back to you once I have something.

fedefernandez:
I'm afraid I don't have enough context to fix the conflicts. Therefore, I vote to close the PR.

PR #588, opened 2023-12-15
AI DSL for Classes, functions and enums

raulraja:
This PR introduces significant improvements in how the Large Language Model (LLM) is used within Kotlin classes, functions, and enums, greatly simplifying its integration and use. The changes focus on automatic inference capabilities, streamlining the process of interfacing with the LLM.

#### Changes Overview:

1. **Classes Enhancement**:
   - Adds a synthetic factory method in the companion objects of classes, allowing for the creation of class instances directly from LLM prompts. This addition simplifies the instantiation process, making it more intuitive and efficient.
   - Example: `Book("A book about hobbits and rings")` directly creates a `Book` instance with relevant properties filled in.

2. **Simplified Function Implementation**:
   - Functions can now be implemented with a straightforward call to `AI("prompt")`. This change allows developers to define function bodies in a more concise manner, leveraging the LLM's capabilities to process and return the expected output.
   - Example: `suspend fun add(x: Int, y: Int): Int = AI("Add $x + $y")` demonstrates this simplified approach.

3. **Optimized Enum Handling**:
   - Introduces an efficient single-token communication method with the LLM for enums. This optimization ensures that only enum cases, which can be encoded as a single token, are accepted. This approach not only speeds up requests to the LLM but also enhances its use in quick decision-making and classification tasks.
   - Example: `Sentiment("I love Xef!")` quickly evaluates the sentiment as `positive`, demonstrating the efficiency of this method.

#### Benefits:

- **Ease of Use**: The PR significantly lowers the barrier to integrating and using the LLM in various Kotlin constructs.
- **Performance Improvement**: Especially in the case of enums, the single-token communication method speeds up LLM interactions.
- **Code Simplification**: Reduces boilerplate and complexity in function and class definitions, making the code more readable and maintainable.


javipacheco (approved):
🚀

PR #598, opened 2023-12-19
Fix response streaming using HttpStatement

raulraja:
The streamed response was fully consumed in memory first because of the call to `response.body`. Instead, we use the recommended ktor docs:

https://ktor.io/docs/response.html#streaming

> When you call the HttpResponse.body function to get a body, Ktor processes a response in memory and returns a full response body. If you need to get chunks of a response sequentially instead of waiting for the entire response, use HttpStatement with scoped [execute](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.statement/-http-statement/execute.html) block.

Additionally, the stream was being folded before emission in the `Chat` interface, which a completion handler and a StringBuilder have replaced to accumulate the final assistant response.

This PR includes also a fix for conversations by default using vector stores that are empty

Run included `SimpleStreaming` example to see the correct behavior of each word being streamed as the AI replies.

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/conversation/Conversation.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/conversation/Conversation.kt
@@ -1,7 +1,6 @@
 package com.xebia.functional.xef.conversation
 
 import com.xebia.functional.openai.apis.ChatApi
-import com.xebia.functional.openai.apis.EmbeddingsApi
 import com.xebia.functional.openai.apis.ImagesApi
 import com.xebia.functional.openai.models.CreateChatCompletionRequestModel
 import com.xebia.functional.openai.models.CreateImageRequest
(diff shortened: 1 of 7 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/conversation/Conversation.kt
```diff
@@ -90,14 +90,15 @@ constructor(
   ): Flow<StreamedFunction<A>> = chat.promptStreaming(prompt, this@Conversation, serializer())
 
   @AiDsl
-  fun promptStreaming(
+  suspend fun promptStreaming(
```
nomisRev, line 93: Why is this `suspend` needed? That should be captured in the `Flow`, no?

raulraja: not needed, thanks!

javipacheco (approved):
🚀

PR #599, opened 2023-12-21
Fix functionCall and toolCalls properties not aligning with API 

raulraja:


openai-client/client/src/commonMain/kotlin/com/xebia/functional/openai/infrastructure/ApiClient.kt
```diff
@@ -35,6 +35,7 @@ open class ApiClient(val baseUrl: String) {
       {
         it.install(ContentNegotiation) { json(jsonBlock) }
         it.install(HttpTimeout)
+        it.install(Logging) { level = LogLevel.NONE }
```
nomisRev, line 38: Should that be configurable? 🤔 Not blocked for this PR.

nomisRev (approved):
Looks good to me! 🙌

javipacheco (approved):
🚀

openai-client/client/src/commonMain/kotlin/com/xebia/functional/openai/infrastructure/ApiClient.kt
javipacheco: Good catch! Yes, this should be configurable. We are going to address this in another PR because this PR is blocking other stuff

PR #600, opened 2023-12-21
Fixes OpenTelemetry tracing scoping

nomisRev:
This PR fixes the current problem, where we see Childs being created as child of each other instead of respecting the correct root.

```
- root span
|-- child span 1
  |-- child span 2
    |-- child span 3
```

This issue was caused by using mutation, and picking the last created span as the parent. This implementation suffers from some other issues like race conditions in the face of parallelism.

To fix this we leverage the created coroutines, and we make each Span in its own coroutine and thus all children are created as a child coroutine of the outer parent span. This fixes the hierarchy:

```
- root span
|-- child span 1
|-- child span 2
|-- child span 3
```

As a side effect `event` and `setAttribute` had to become `suspend` since we need to extract the OpenTelemetry `Context` from the `CoroutineContext` but this was non-problematic because all the surrounding code already supports `suspend`.


raulraja (approved):
thanks @nomisRev !

Montagon (approved):
🚀

franciscodr (approved):
Excellent! Thanks, @nomisRev!

PR #601, opened 2023-12-21
Check local embedding before calling embeddings api

javipacheco:
This PR checks if the database contents embeddings before calling OpenAI to avoid extra calls to the model


franciscodr (approved):
I have tested it locally in an application and it works fine

Montagon (approved):

PR #602, opened 2023-12-24
AI DSL : Generic serialization support for Sealed classes and other primitives

raulraja:
This Pull Request introduces support for serializing sealed classes in Kotlin Serialization. It does so by mapping function names to the serial names of descriptors and integrating them into the JSON object processed by Kotlin Serialization. This enhancement enables the sealed class serializer to correctly identify and handle the appropriate class.

Here's an example demonstrating the support for sealed classes:

```kotlin
package com.xebia.functional.xef.dsl

import com.xebia.functional.xef.AI
import kotlinx.serialization.Serializable

@Serializable
sealed class Response {

  @Serializable
  data class City(val name: String) : Response()

  @Serializable
  data class Country(val name: String) : Response()

  @Serializable
  data class Continent(val name: String) : Response()
}

suspend fun main() {
  val response = AI<Response>("Capital of France?")
  println(response) // This will print: City(name=Paris)
}
```

In this example, the `Response` sealed class has three subclasses: `City`, `Country`, and `Continent`. Depending on the question asked, the appropriate subclass is instantiated and serialized.

Additionally, the PR extends support to primitives and other types that are not wrapped in a `JsonObject`. Here's an example:

```kotlin
suspend fun main() {
  val two: Int = AI("What is 1 + 1?")
  val truth: Boolean = AI("Is the earth flat?")
  val name: String = AI("Hi AI, What is your name?")
  println(
    """
    |two: $two
    |truth: $truth
    |name: $name
    """.trimMargin()
  )
}
```

In this second example, simple types like `Int`, `Boolean`, and `String` are used. The AI function returns the appropriate type based on the question asked, demonstrating the flexibility and utility of the new serialization support.


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/ChatWithFunctions.kt
```diff
@@ -31,8 +36,29 @@ suspend fun <A> ChatApi.prompt(
...
+): A =
+  prompt(prompt, scope, chatFunctions(descriptors)) { call ->
+    // adds a `type` field with the call.functionName serial name equivalent to the call arguments
+    val jsonWithDiscriminator = Json.decodeFromString(JsonElement.serializer(), call.arguments)
+    val descriptor =
+      descriptors.firstOrNull { it.serialName.endsWith(call.functionName) }
+        ?: error("No descriptor found for ${call.functionName}")
+    val newJson =
+      JsonObject(
+        jsonWithDiscriminator.jsonObject + ("type" to JsonPrimitive(descriptor.serialName))
```
raulraja, line 59: This addition supports polymorphic types like sealed classes whose serializer is chosen by the `type` property, the discriminator.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/ChatWithFunctions.kt
```diff
@@ -92,43 +95,57 @@ suspend fun <A> ChatApi.prompt(
           requestedMemories,
           prompt.configuration.messagePolicy.addMessagesToConversation
         )
-        .mapNotNull { it.message.toolCalls?.firstOrNull()?.function?.arguments }
+        .mapNotNull {
+          val functionName = it.message.toolCalls?.firstOrNull()?.function?.name
```
raulraja, line 99: We also need the name to find the correct descriptor from the cases of the sealed class.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
```diff
@@ -21,6 +21,9 @@ suspend fun <T> Prompt<T>.addMetrics(conversation: Conversation) {
     "${messages.size} (${messages.map { it.completionRole().value.firstOrNull() ?: "" }.joinToString("-")})"
   )
   conversation.metric.parameter("conversation-id", conversation.conversationId?.value ?: "none")
-  conversation.metric.parameter("functions", function?.let { "yes" } ?: "no")
+  conversation.metric.parameter(
+    "functions",
+    if (functions.isEmpty()) "no" else functions.joinToString(",") { it.name }
```
raulraja, line 26: Instead of tracking `yes` we track the actual function names

openai-client/client/src/commonMain/kotlin/com/xebia/functional/openai/models/ext/chat/ChatCompletionToolChoiceOption.kt
```diff
@@ -0,0 +1,42 @@
...
+ * message or calling a function. Specifying a particular function via `{\"type: \"function\",
+ * \"function\": {\"name\": \"my_function\"}}` forces the model to call that function. `none` is the
+ * default when no functions are present. `auto` is the default if functions are present.
+ *
+ * @param type The type of the tool. Currently, only `function` is supported.
+ * @param function
+ */
+@JvmInline
+@Serializable
+value class ChatCompletionToolChoiceOption(val element: JsonElement) {
```
raulraja, line 25: This is a trick to support the spec supporting either String or a complex json object that specifies a function name.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/ChatWithFunctions.kt
```diff
@@ -31,8 +36,29 @@ suspend fun <A> ChatApi.prompt(
...
+  prompt: Prompt<CreateChatCompletionRequestModel>,
+  scope: Conversation,
+  serializer: KSerializer<A>,
+  descriptors: List<SerialDescriptor>,
+): A =
+  prompt(prompt, scope, chatFunctions(descriptors)) { call ->
+    // adds a `type` field with the call.functionName serial name equivalent to the call arguments
+    val jsonWithDiscriminator = Json.decodeFromString(JsonElement.serializer(), call.arguments)
+    val descriptor =
+      descriptors.firstOrNull { it.serialName.endsWith(call.functionName) }
```
raulraja, line 55: This can be an issue because we use simple function names without FQN. Sadly, sending the full package name to Open AI is not a good idea because selecting the proper function makes it more challenging. We bail if the function is not found in the cases, forcing the user to reencode it or provide `@SerialName`.

Montagon (dismissed):
It's great @raulraja!! 🔝

Montagon (approved):

PR #609, opened 2023-12-26
Add seed parameter in PromptConfiguration

Montagon:
This PR adds the support of the `seed` parameter inside the `PromptConfiguration`.


raulraja (approved):
thanks @Montagon

PR #623, opened 2024-01-03
Span as client

franciscodr:
This pull request introduces changes in how the span's attributes are represented within the OpenTelemetry traces. We have followed some of the [semantic conventions](https://opentelemetry.io/docs/specs/semconv/) included on the Opentelemetry website.

The proposed changes are:
- Split the information about the consumed tokens into three different attributes:
  - Prompt
  - Completion
  - Total
- Set the kind of the spans as `client` to represent better what the interactions with OpenAI do. OpenTelemetry documentation says about the client spans:
  > A client span represents a synchronous outgoing remote call such as an outgoing HTTP request or database call.

- Rename some of the span attributes to follow naming conventions
- Add a new function that allows setting an array of strings as the value of a span attribute.
- Group all the attributes under the `openai` namespace

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
@@ -17,14 +17,16 @@ suspend fun CreateChatCompletionResponse.addMetrics(
 }
 
 suspend fun <T> Prompt<T>.addMetrics(conversation: Conversation) {
+  conversation.metric.parameter("openai.chat_completion.prompt.message.count", "${messages.size}")
+
   conversation.metric.parameter(
-    "openai.chat_completion.prompt.message.count",
-    "${messages.size} (${messages.map { it.completionRole().value.firstOrNull() ?: "" }.joinToString("-")})"
+    "openai.chat_completion.prompt.messages",
+    messages.map { it.completionRole().value }
   )
-  conversation.metric.parameter("openai.chat_completion.conversation.id", conversation.conversationId?.value ?: "none")
+  conversation.metric.parameter("openai.chat_completion.conversation_id", conversation.conversationId?.value ?: "none")
   conversation.metric.parameter(
     "functions",
-    if (functions.isEmpty()) "no" else functions.joinToString(",") { it.name }
+    if (functions.isEmpty()) listOf("no") else functions.map { it.name }
   )
   conversation.metric.parameter("openai.chat_completion.prompt.temperature", "${configuration.temperature}")
 }
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
@@ -10,7 +10,10 @@ suspend fun CreateChatCompletionResponse.addMetrics(
   conversation.metric.parameter("openai.chat_completion.model", model)
   usage?.let {
     conversation.metric.parameter("openai.chat_completion.prompt.token.count", "${it.promptTokens}")
-    conversation.metric.parameter("openai.chat_completion.completion.token.count", "${it.completionTokens}")
+    conversation.metric.parameter(
+      "openai.chat_completion.completion.token.count",
+      "${it.completionTokens}"
+    )
     conversation.metric.parameter("openai.chat_completion.token.count", "${it.totalTokens}")
   }
   return this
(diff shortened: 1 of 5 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
```diff
@@ -7,23 +7,39 @@ import com.xebia.functional.xef.prompt.Prompt
...
+  conversation.metric.parameter(
+    "openai.chat_completion.conversation_id",
+    conversation.conversationId?.value ?: "none"
+  )
   conversation.metric.parameter(
-    "number-of-messages",
-    "${messages.size} (${messages.map { it.completionRole().value.firstOrNull() ?: "" }.joinToString("-")})"
+    "openai.chat_completion.functions",
+    if (functions.isEmpty()) listOf("no") else functions.map { it.name }
   )
```
nomisRev, line 40: **Nit**, wondering if we should rename `no` to `none` or if we should omit `"openai.chat_completion.functions"` if `functions.isEmpty()`.

So:

```
if(functions.isNotEmpty()) {
  conversation.metric.parameter(
    "openai.chat_completion.functions",
     functions.map { it.name }
  )
}
```

core/src/commonMain/kotlin/com/xebia/functional/xef/metrics/LogsMetric.kt
```diff
@@ -46,5 +46,9 @@ class LogsMetric : Metric {
     logger.info { "${writeIndent(numberOfBlocks.get())}|-- $key = $value" }
   }
 
+  override suspend fun parameter(key: String, values: List<String>) {
+    logger.info { "${writeIndent(numberOfBlocks.get())}|-- $key = $values" }
```
nomisRev, line 50: Should we mention here that it's about a `parameter` in the log message?

**off-topic:** should we take a `LogLevel` as a parameter in `LogsMetric`?

integrations/opentelemetry/src/main/kotlin/com/xebia/functional/xef/opentelemetry/OpenTelemetryState.kt
```diff
@@ -34,4 +37,11 @@ class OpenTelemetryState(val tracer: Tracer) {
     Context.current().with(span)
     span.setAttribute(key, value)
   }
+
+  suspend fun setAttribute(key: String, values: List<String>) {
+    val span = currentCoroutineContext().getOpenTelemetryContext().let(Span::fromContext)
+
+    Context.current().with(span)
```
nomisRev, line 44: I think I made a mistake in `setAttribute`, that was copied here:

`currentCoroutineContext().getOpenTelemetryContext()` returns the `Context` that is held in the `CoroutineContext`, so `Current.context().with(span)` seems incorrect here 🤔.

  => `Context.current()` works with `ThreadLocal`, and `currentCoroutineContext().getOpenTelemetryContext()` uses the `CoroutineContext`.

What we're doing in seems to be getting the `Span` from the `Context` in the `currentCoroutineContext()` in ` currentCoroutineContext().getOpenTelemetryContext().let(Span::fromContext)`.

And then setting that `Span` in the `Context.current()`, so the `ThreadLocal` one. There should be no need for that at all.

The goal afaik, is just to get the current `Span` from the current Context `currentCoroutineContext().getOpenTelemetryContext()`, and set an attribute to it.

So this code should be sufficient:

```kotlin
currentCoroutineContext()
  .getOpenTelemetryContext()
  .let(Span::fromContext)
  .setAttribute(AttributeKey.stringArrayKey(key), values)
```

nomisRev (commented):
I think there was some old code of mine left in `setAttribute`, I left some comments to clarify and explain.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
franciscodr: I like this approach.

integrations/opentelemetry/src/main/kotlin/com/xebia/functional/xef/opentelemetry/OpenTelemetryState.kt
franciscodr: Good point. Yeah, I don't think we need to call `Context.current()` in the code anymore.

core/src/commonMain/kotlin/com/xebia/functional/xef/metrics/LogsMetric.kt
franciscodr: Let me think about it. We could probably address this in a separate pull request.

nomisRev:
@franciscodr I think we agree on all points. If the fixes/simplifications are added then I will approve the PR. I can create the irrelevant change in another PR ☺️

nomisRev (approved):
Awesome, thank you @franciscodr

franciscodr:
Thanks for the review, @nomisRev 😊

PR #631, opened 2024-01-10
Assistants API: Do not attempt to call functions without arguments

raulraja:
Often, the assistant API returns empty arguments for functions that have required arguments.
This check ensures we only attempt to process functions without arguments.


javipacheco (approved):
🚀

PR #632, opened 2024-01-10
Remove the openai-client dependency from the project

franciscodr:
We are no longer using this dependency in our code:
* `com.aallam.openai:openai-client`

So this pull request removes it from the dependencies section in the `server` project.


javipacheco (approved):
🚀

Montagon (approved):

PR #644, opened 2024-01-24
New Json instance in ChatWithFunctions

javipacheco:


raulraja (changes requested):
This work should happen in the ApiClient mustache template which is where the default JSON is configured.
https://github.com/xebia-functional/xef/blob/main/openai-client/generator/config/libraries/multiplatform/infrastructure/ApiClient.kt.mustache#L75-L79

raulraja (approved):

PR #646, opened 2024-01-26
Avoiding extra calls in ToolCalls

javipacheco:
We have detected when we are emitting messages in the stream, we are sending extra calls. That happens when the `status` is `in_progress`  and `type` is `tool_calls`. Sometimes the tools don't have any arguments (maybe OpenAI is preparing the information to send to the tool) and we have to ignore this type of events


raulraja (approved):
thanks @javipacheco !

PR #696, opened 2024-03-21
Update/testsuite

Montagon:


raulraja (dismissed):

raulraja (approved):

javipacheco (approved):
The proposal of the tests should be if the response is successful or not. In this case, we should be able to say what results of the metrics are the positives in order to say if the test failed. 

We can address this in another separate PR. Anyway, I think that we should change the `unaryPlus` in the DSL

Good job!

javipacheco (approved):
🚀

PR #698, opened 2024-03-23
Update OpenAI model types

TheDome0:
Add GPT_3_5_TURBO_0125 and update default GPT_3_5_TURBO context window according to https://platform.openai.com/docs/models/gpt-3-5-turbo


raulraja:
Hi @TheDome0 
Seems to be failing with:
```
com.xebia.functional.xef.conversation.ConversationSpec[jvm] > " | GPT_3_5_TURBO model has 4097 max context length | when the number of token in the conversation is greater than | the space allotted for the message history in the prompt configuration | the number of messages in the request must have fewer messages than | the total number of messages in the conversation |[jvm] FAILED
    java.lang.AssertionError: 100 should be < 100
        at com.xebia.functional.xef.conversation.ConversationSpec$1$2.invokeSuspend(ConversationSpec.kt:88)

        Caused by:
        java.lang.AssertionError: 100 should be < 100
            at com.xebia.functional.xef.conversation.ConversationSpec$1$2.invokeSuspend(ConversationSpec.kt:88)
```
That test may need to get updated at https://github.com/xebia-functional/xef/blob/main/core/src/commonTest/kotlin/com/xebia/functional/xef/conversation/ConversationSpec.kt#L54

TheDome0:
Well, shouldn't the check simply be the same as for the 16k model right below? Since now the default 3.5 model also has a 16k context length.
So instead of `messagesSizePlusMessageResponse shouldBeLessThan memories.size` it should be `messagesSizePlusMessageResponse shouldBe memories.size`

PR #707, opened 2024-03-28
Enable Assistants Streaming

raulraja:
Enable assistant message deltas and streaming for other endpoints based on https://platform.openai.com/docs/api-reference/assistants-streaming


nomisRev (dismissed):

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/AssistantThread.kt
```diff
@@ -72,208 +69,110 @@ class AssistantThread(
...
+                run.status == RunObject.Status.requires_action &&
+                  run.requiredAction?.type == RunObjectRequiredAction.Type.submit_tool_outputs
+              ) {
+                val callsResult: List<Pair<String, Assistant.Companion.ToolOutput>> =
+                  calls
+                    .filterIsInstance<
+                      RunStepDetailsToolCallsObjectToolCallsInner.CaseRunStepDetailsToolCallsFunctionObject
+                    >()
+                    .parMap { toolCall -> executeToolCall(toolCall, assistant) }
+                    .filterNotNull()
```
nomisRev, line 97: @raulraja you added `parMapNotNull` into Arrow couple months ago 😜 We can use that here.
The suggestion will fail with a missing import though.

```suggestion
                    .parMapNotNull { toolCall -> executeToolCall(toolCall, assistant) }
```

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/AssistantThread.kt
```diff
@@ -72,208 +69,110 @@ class AssistantThread(
...
+                val results: Map<String, Assistant.Companion.ToolOutput> = callsResult.toMap()
+                val toolOutputsRequest =
+                  SubmitToolOutputsRunRequest(
+                    toolOutputs =
+                      results.map { (toolCallId, result) ->
```
nomisRev, line 102: We can potentially also accumulate multiple json parsing errors if you want. I can add it if you like. Something like:

```
listOf("on", "tw", "thre")
    .mapOrAccumulate {
        Either.catch { it.toInt() }.bind()
    } // fold into single Throwable, and rethrow?
```

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/RunDeltaEvent.kt
```diff
@@ -0,0 +1,27 @@
+package com.xebia.functional.xef.llm.assistants
+
+enum class RunDeltaEvent {
```
nomisRev, line 3: Should this be private inside of `RunDelta.kt`? Or is there a need to export this?

openai-client/generator/config/model_oneof.mustache
```diff
@@ -24,10 +24,11 @@ object {{classname}}Serializer : KSerializer<{{classname}}> {
     }
 
     override fun deserialize(decoder: Decoder): {{classname}} {
+      val jsonSerde = Json { ignoreUnknownKeys = true }
```
nomisRev, line 27: Hmm. Should we go back to a global instance instead? It's meant for OpenAI specifically anyway here, and recreating the `Json` is quite expensive (I think it initialises some reflection stuff).

Can also take care of this in a subsequent PR, there is probably more cases where we want to change hardcoded instances of `Json`.

nomisRev (approved):
Looks great @raulraja. Only couple of nits, but they can be addressed afterwards.

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/AssistantThread.kt
raulraja: In this case, so far, it's not needed because if a tool fails with an exception, we turn that exception into a json object we send to the LLM to reason about the output.
https://github.com/xebia-functional/xef/blob/dc55080b07f649d11bff74051e62bea7c136dc33/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/Assistant.kt#L50-L70

openai-client/generator/config/model_oneof.mustache
raulraja: Yes, we need to go back to a global instance, which is probably better done in another PR. ChatWithFunctions also uses a new instance each time for deserializing the LLM function call, and sometimes, it fails because it's not permissive enough. That one also needs config for `ignoreUnknownKeys`.

We also noticed the serializers are using the `type` property defined twice in objects with a property with the same name. We may have to change the `classDiscriminator` globally so it is not called `type` in the Json object config for all calls.

javipacheco (approved):
🚀  LGTM!

I have tested this PR locally and everything works as expected

openai-client/generator/config/model_oneof.mustache
raulraja: merging this now, let's do this in another PR

PR #718, opened 2024-04-12
Recurse if after `tool submission` `run` still requires action.

raulraja:
This pull request refactors the `AssistantThread` class to handle better scenarios where a run still requires action after tool output submissions. The main changes include introducing a `takeRequiredAction` method that manages tool output submission, and checks run status recursively. This ensures all necessary actions are processed before a run is considered completed.


javipacheco (approved):
LGTM! 🚀 

I've tested locally and everything works fine

PR #756, opened 2024-06-06
OpenTelemetry improved

javipacheco:
This PR:

- Adds the missing elements in the stream
- Write the name of the data class in the span instead of the status
- Calculate the time os the span in the Run, RunStep and MessageCreation

Screenshot 2024-06-06 at 16 41 27

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
@@ -1,6 +1,7 @@
 package com.xebia.functional.xef.llm
 
 import com.xebia.functional.openai.generated.model.CreateChatCompletionResponse
+import com.xebia.functional.openai.generated.model.MessageObject
 import com.xebia.functional.openai.generated.model.RunObject
 import com.xebia.functional.openai.generated.model.RunStepObject
 import com.xebia.functional.xef.conversation.Conversation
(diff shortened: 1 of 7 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
```diff
@@ -60,33 +61,42 @@ suspend fun Prompt.addMetrics(conversation: Conversation) {
...
+    is RunDelta.RunStepCompleted -> runStep.addMetrics(metric, "RunStepCompleted")
+    is RunDelta.RunStepCreated -> runStep.addMetrics(metric, "RunStepCreated")
+    is RunDelta.RunStepExpired -> runStep.addMetrics(metric, "RunStepExpired")
+    is RunDelta.RunStepFailed -> runStep.addMetrics(metric, "RunStepFailed")
+    is RunDelta.RunStepInProgress -> runStep.addMetrics(metric, "RunStepInProgress")
+    is RunDelta.MessageCreated -> message.addMetrics(metric, "MessageCreated")
+    is RunDelta.MessageIncomplete -> message.addMetrics(metric, "MessageIncomplete")
+    is RunDelta.MessageCompleted -> message.addMetrics(metric, "MessageCompleted")
+    is RunDelta.MessageInProgress -> message.addMetrics(metric, "MessageInProgress")
     else -> {} // ignore other cases
```
Montagon, line 100: do you think it will be interesting to add all the remaining cases and then, do nothing?

```kotlin
is RunDelta.MessageInProgress -> message.addMetrics(metric, "MessageInProgress")
is RunDelta.AnotherOne,
is RunDelta.Extra -> {}
```

In that case, if there is a new case, this will fail and we can check if we need to add new metric for this new case.

Montagon (dismissed):
Just a comment, thanks Javi!

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/MetricManagement.kt
javipacheco: Good catch!

raulraja (approved):

PR #757, opened 2024-06-11
Cached Tool

javipacheco:
This PR creates a new `CachedTool`

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/CachedTool.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/CachedTool.kt
@@ -6,7 +6,7 @@ import kotlin.time.Duration
 import kotlin.time.Duration.Companion.days
 
 abstract class CachedTool<Input, Output>(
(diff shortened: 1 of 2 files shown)


Montagon (dismissed):

core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/CachedTool.kt
```diff
@@ -0,0 +1,33 @@
+package com.xebia.functional.xef.llm.assistants
+
+import arrow.fx.coroutines.Atomic
+import arrow.fx.coroutines.timeInMillis
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.days
+
+abstract class CachedTool<Input, Output>(
+  private val cache: Atomic<MutableList<CachedToolInfo<Input, Output>>>,
```
raulraja, line 9: Why not a `MutableMap<Input, CachedToolInfo<Input, Output>>`? That way, the lookups by `Input` would be faster and would solve the issue of allowing duplicates, which currently the list does.

raulraja (approved):

Montagon (approved):

PR #775, opened 2024-08-14
Expose cache as a map of filtered seed

realdavidvega:
# This PR
- Creates a `getCache` function in `CachedTool`, which inside an atomic operation removes the expired entries for the instance `seed`, modifies the state of the map and returns a map containing only the not expired entries for the instance seed.
- Refactors `cache` function to use equivalent indexing instead of calling.


javipacheco (approved):
:rocket:

PR #787, opened 2024-09-11
Add support for gpt-4o-mini in ModelType

realdavidvega:
# Background

`ModelType` didn't have the enum of `gpt-4o-mini` with the right encoding configuration, leading to falling back to `ModelType.TODO`, which has `CL100K` encoding. 

This led to errors when using the tokenizer decoder, like in the `AI.runWithEnumSingleTokenSerializer`, which was used for one token enums.

# This PR 

Adds `ModelType.GPT_4O_MINI` with `O200K_BASE` encoding and `128000` context length.

Example:
```kotlin
@Serializable
enum class Sentiment {
  positive,
  negative,
  neutral
}

suspend fun main() {
  val sentiment = AI<Sentiment>("I love Xef!", config = AIConfig(model = CreateChatCompletionRequestModel.gpt_4o_mini))
  println(sentiment) // positive
}
```

which returns `positive`.


Montagon (approved):
🚀

PR #795, opened 2024-10-10
Handle OpenAI Error server event accordingly

franciscodr:
We are currently handling the `Error` server events coming from OpenAI as `Unknown`, preventing us from getting more accurate information about what went wrong during the interaction with the OpenAI API.

This pull request handles the errors accordingly, decoding the incoming JSON as an `Error` class instance. This change will allow us to respond better to the final user later.

### DIFF
--- a/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/RunDelta.kt
+++ b/core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/RunDelta.kt
@@ -134,15 +134,22 @@ sealed interface RunDelta {
   /** [RunDeltaEvent.thread_message_incomplete] */
   @JvmInline @Serializable value class MessageIncomplete(val message: MessageObject) : RunDelta
 
+  /** [RunDeltaEvent.error] */
+  @JvmInline
+  @Serializable
+  value class Error(val error: com.xebia.functional.openai.generated.model.Error) : RunDelta
+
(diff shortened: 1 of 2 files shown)


core/src/commonMain/kotlin/com/xebia/functional/xef/llm/assistants/RunDelta.kt
```diff
@@ -134,15 +134,22 @@ sealed interface RunDelta {
...
+    private val typeExcludedCharactersRegex = "[._]".toRegex()
+
     fun fromServerSentEvent(serverEvent: ServerSentEvent): RunDelta {
       val data = serverEvent.data ?: error("Expected data in ServerSentEvent for RunDelta")
       val type = serverEvent.event ?: error("Expected event in ServerSentEvent for RunDelta")
       val event =
-        RunDeltaEvent.values().find {
-          type.replace(".", "").replace("_", "").equals(it.name, ignoreCase = true)
+        RunDeltaEvent.entries.find {
+          type.replace(typeExcludedCharactersRegex, "").equals(it.name, ignoreCase = true)
```
franciscodr, line 152: Remove the deprecated `values()` call and use the recommended `entries` property. It also replaces chained `replace` calls with a regular expression containing all the characters we want to remove.

javipacheco (approved):

realdavidvega (approved):
LGTM!

PR #796, opened 2024-10-11
Better serializer errors

realdavidvega:
## Background
In the current implementation, the error handling mechanism for HTTP responses is simplistic and makes more difficult for consumers to effectively identify and handle those errors.

## This PR

Introduces a structured `ResponseErrorInfo` data class that encapsulates all relevant details of an HTTP transaction, including request URL, method, and body, as well as response status, headers, and body. Also modifies `ResponseSerializerError` to include this information.


Montagon (approved):

javipacheco (approved):
🚀

